PR #9, opened 2016-07-28
add session checking endpoints to nodeJs and call session check every…

ahsu1230:
… 10 seconds


ahsu1230:
lodash is updated to the full build since it was missing a lot of useful functions

PR #12, opened 2016-07-30
Add FileClaim model, service, controller + refactor

ahsu1230:


winterfell/controllers/authController.js
```diff
@@ -41,17 +42,17 @@ module.exports = function (app) {
...
           onError: function(errorCode, errorMsg) {
             console.log('[authSignUp] ' + errorCode + ' Error creating user: ' + errorMsg);
-            res.sendStatus(errorCode, errorMsg);
+            res.status(errorCode).send({error: errorMsg});
           },
           onSuccess: function(user) {
             console.log('[authSignUp] Successfully created user ' + user.externalId);
             var extUserId = user.externalId;
-            res.status(200).send({userId: extUserId, redirect: '/'});
+            res.status(http.CODE_SUCCESS).send({userId: extUserId, redirect: '/'});
```
qwwqwwq: I think its a little confusing to have nonstandard names for the HTTP status codes. Fine with having constants for them, but if we do I think the constant for 200 should be name "OK" etc.

winterfell/utils/httpResponses.js
```diff
@@ -0,0 +1,33 @@
+function httpResponses(app) {
+    this.app = app;
+};
+
+module.exports = httpResponses;
+
+// Http Codes (wrapper around standard http response status codes)
```
qwwqwwq: Can we just use https://www.npmjs.com/package/http-status-codes instead of rolling this ourselves?

ahsu1230: Ah ok, I would like to have standardized error message constants so I'll keep that for now

ahsu1230:
I'll keep these changes smaller in the future.
Sorry, just wanted to refactor to make things neater.

PR #27, opened 2016-08-05
Adds Formly

qwwqwwq:
Formly is a library for creating forms in angular from a json schema, it seems very flexible and easy to extend, and I think it will work better for our use case than json-object-schema, the previous library we used for this


winterfell/webapp/gulpfile.js
```diff
@@ -97,4 +102,6 @@ gulp.task('watch', ['build', 'initBrowserSync'], function () {
 
 gulp.task('build', ['clean', 'fonts', 'icons', 'libs', 'js', 'other-js', 'stylus', 'jade']);
 
-gulp.task('default', ['clean', 'build', 'initBrowserSync', 'watch']);
+gulp.task('default', ['clean', 'build'
```
ahsu1230: if nothing is being added here, could you change it back?

winterfell/webapp/gulpfile.js
```diff
@@ -77,7 +81,8 @@ gulp.task('fonts', function() {
 
 gulp.task('icons', function(){
   return gulp.src('src/icons/**/*.+(png|jpg|gif|svg)')
-  .pipe(cache(imagemin()))
+  //.pipe(debug({title: 'icons:'}))
```
ahsu1230: removed this part - no more imagemin()

winterfell/webapp/src/js/formCtrl.js
```diff
@@ -9,6 +9,29 @@ app.controller("formCtrl", ['$scope', 'formRenderingService', function($scope, f
...
     $scope.render = function() {
-        formRenderingService.render('VBA-21-0966-ARE', [{fieldName: 'veteran_first_name', fieldValue: 'jeff'}]);
-    }
+        var out = [];
+        for (var k in $scope.model) {
+            if ($scope.model.hasOwnProperty(k)) {
+                out.push({fieldName: k, fieldValue: $scope.model[k]})
+            }
+        }
+        formRenderingService.render('VBA-21-0966-ARE', out);
```
ahsu1230: it would be good practice to start placing constants like Form names, titles, fields, etc. into its own "class".
Feel free to use whatever makes sense for you.
I have a strings.js (containing military ranks -> text, not that we really need it, but it's there) in the `libs` folder and its gets treated like a library so you can to import it elsewhere.

winterfell/webapp/src/js/formCtrl.js
```diff
@@ -9,6 +9,29 @@ app.controller("formCtrl", ['$scope', 'formRenderingService', function($scope, f
...
+    };
+
+    $scope.onSubmit = function() {
+        $scope.render();
+    };
+
+    $scope.fields = [
+        {
+            key: 'veteran_first_name',
```
ahsu1230: same with form names

ahsu1230:
👍  other than view nits and merge conflict!

winterfell/webapp/src/js/formCtrl.js
qwwqwwq: Agree just had this in there as I am developing, planning to get it from route params eventually, or maybe have a mapping from route params to constants

PR #30, opened 2016-08-08
Removes imagemin

qwwqwwq:


ahsu1230:
👍 Locally, I'll try to figure out how much we're "saving" to see if imagemin made a difference anyway.

PR #31, opened 2016-08-08
Style selecting forms page for a claim

ahsu1230:
- Added new icons (a next arrow + docs icon)
- Select Forms now gets a link that forwards you to a form at `/claim/form/:formId`. The idea is that after you fill the form and go back, you can click to fill another form with a different `id`.
- Footer gets a CANCEL button or FINISH button. Canceling will remove the claim state. Finishing will bring you to the claim confirm page.
- Forms also get an extra attribute: required and completed. The select-forms page will respond differently based on whether a form is required and not filled or completed. Soon, we can look into how partially (not finished) forms can look.
- Small things like replacing the home Docs icon with a different one.


ahsu1230:
@qwwqwwq

qwwqwwq:
LGTM

PR #32, opened 2016-08-09
Setup Dynamic Form Templating

qwwqwwq:
- Adds initial form template for intent to file
- Adds static serving of the formly templates, and complimentary angular service.
- Changes /form path to include routeParam that determines which form is displayed & rendered.


winterfell/webapp/src/js/formRenderingService.js
```diff
@@ -27,3 +27,29 @@ app.factory('formRenderingService', ['$window', '$http', function ($window, $htt
         }
     }
 }]);
+
+/**
+ * Service to retrieve the formly template (which is just a static JSON file)
+ * for a given form.
+ */
+app.factory('formTemplateService', ['$http', function ($http) {
```
ahsu1230: There is a service called net.js which handles API endpoints to our server.
You should use that if you want to make HTTP requests

ahsu1230: it adds a wrapper around $http so all you have to provide is a url and body (if needed)

ahsu1230:
other than one comment, looks fine to me

winterfell/webapp/src/js/formRenderingService.js
qwwqwwq: Which library is that, I don't see anything called net in the package json? I agree $http is a little kludgy, I think using $resource might be preferred.

ahsu1230: it's not in package.json. It's a custom file on the same level as this file 'webapp/src/js/net.js'.

qwwqwwq:
Decided not to use xhr to get the form schemas, it is a little slow and the schemas are not that big

PR #33, opened 2016-08-09
Server form model

ahsu1230:
Here is a new `form` model that will handle:

- generic metadata about some form
- what fields exist in form
- which fields are required for form to be complete
- is form required for ALL claims

To help with this, we have a helper class FormProperties that contains constants about `form_id` and `form_fields`.

For more info, please refer to this google doc:
https://docs.google.com/document/d/1OxwnoH7X1JyjI9-0-qsE6GSyd5oqbmNjFKWFtJDfkTU/edit?usp=sharing


ahsu1230:
screwed up the merging from before

winterfell/config/mongoose.js
```diff
@@ -13,6 +18,7 @@ module.exports = function(environment) {
   });
   db.on('open', function () {
       console.log('Database connection opened at ' + address);
+      setupDb();
```
qwwqwwq: So our database is going to have a new record inserted each time we startup the app? Eventually we will have a lot of duplicate records in the DB right?

winterfell/models/form.js
```diff
@@ -0,0 +1,48 @@
...
+  required_fields: [String],        // Array of "form fields" that Form requires
+  required: Boolean                 // Is form required for every claim?
+});
+
+// Query helpers (none)
+// Instance methods (none)
+
+// Static methods
+var allForms = [
+  {
```
qwwqwwq: I dont see the benefit of putting our configs in the db like this, the form configurations should be more accessible. In my other revision I was looking at putting them in static json files that will be accessible to all parts of our app

winterfell/models/form.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+var _ = require('lodash');
+var FormProps = require('../utils/formProperties');
+var mongoose = require('mongoose');
+var Schema = mongoose.Schema;
+
+var FormSchema = new Schema({
+  _id: String,
```
qwwqwwq: _id is a special field in mongodb, with type ObjectID (not String), I'm not sure how mongoose allows this is must be renaming or something on the backend, but we should not make models that override _id, its confusing

winterfell/utils/formProperties.js
```diff
@@ -0,0 +1,54 @@
+function FormProperties(app) {
+    this.app = app;
+}
+
+module.exports = FormProperties;
+
+module.exports.FORM_ID = {
+  INTENT_TO_FILE_CLAIM: 'vba_21_0966_are'
```
qwwqwwq: This type of form configuration information should be a static config file, so it is accessible to non-javascript code

qwwqwwq:
Also I see you are checking in the intent to file form, I have already checked it in at `pdf-filler/src/main/resources/forms/VBA-21-0966-ARE.pdf`, I'm not sure anything other than the pdf microservice cares about the pdf file itself so thats why its buried in there. Did you want to move the forms up higher in the directory tree?

winterfell/config/mongoose.js
ahsu1230: we could have a clearDb() function that either drops the db or at least clears all entries.
Maybe a better option is to only add if it doesn't exist yet.

ahsu1230:
We can keep the pdfs in there. I made the `resources` folder as an easy way to retrieve any important files.

winterfell/utils/formProperties.js
ahsu1230: ok I'll label a `static` folder which has static configurations shared by both backend and frontend.

winterfell/models/form.js
ahsu1230: mongoose does allow renaming of keys, but as you said, form should be a static json

PR #34, opened 2016-08-10
Create UserValue db model

ahsu1230:
@qwwqwwq

### DIFF
--- a/winterfell/models/userValues.js
+++ b/winterfell/models/userValues.js
@@ -1,0 +1,24 @@
+'use strict';
+var mongoose = require('mongoose');
+var User = require('./user');
+var Schema = mongoose.Schema;
+
+var UserValueSchema = new Schema({
+  user_id: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'User'
+  },
+  values: Schema.Types.Mixed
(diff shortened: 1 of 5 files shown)


winterfell/models/userValues.js
```diff
@@ -0,0 +1,24 @@
...
+var User = require('./user');
+var Schema = mongoose.Schema;
+
+var UserValueSchema = new Schema({
+  userId: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'User'
+  },
+  values: Schema.Types.Mixed
+}, { timestamps: true });
```
ahsu1230: with this, Schema will automatically create a `createdAt` and `updatedAt` fields for you.
Because these default values are camelcase and using snakecase for db fields in the first commit was a bit awkward, let's just go ahead and use camelcase all around our casebase.

Maybe at least constants can still be CAPITAL_SNAKE_CASE

qwwqwwq: Agree everything in the javascript world should be camel case, thats kind of the convention everyone follows. Mongo is technically JSON-ish so using camek case in there makes sense.

qwwqwwq:
LGTM

PR #35, opened 2016-08-10
Add timestamps to appropriate models and fixed fileClaimTests

ahsu1230:
@qwwqwwq

### DIFF
--- a/winterfell/controllers/fileClaimController.js
+++ b/winterfell/controllers/fileClaimController.js
@@ -39,18 +39,17 @@ module.exports = function (app) {
   app.post('/claims/create', function (req, res) {
     console.log('[createClaim] request received for ' + JSON.stringify(req.body));
     var extUserId = req.body.extUserId;
-    var callbacks = {
-      onSuccess: function(claim) {
+    var callback = function(err, claim) {
+      if (claim) {
         res.status(http.OK).send({claim: FileClaim});
-      },
-      onError: function(errCode, status) {
+      } else {
         res.status(http.INTERNAL_SERVER_ERROR).send({error: httpErrors.DATABASE});
       }
-    }
+    };
     if (extUserId) {
       User.findOne({externalId: extUserId}).exec(function(err, user) {
         if (user) {
-          FileClaimService.createNewClaim(user.id, callbacks);
+          FileClaimService.createNewClaim(user.id, callback);
         } else {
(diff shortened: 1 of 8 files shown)


winterfell/controllers/fileClaimController.js
```diff
@@ -39,18 +39,17 @@ module.exports = function (app) {
...
+      } else {
         res.status(http.INTERNAL_SERVER_ERROR).send({error: httpErrors.DATABASE});
       }
-    }
+    };
     if (extUserId) {
       User.findOne({externalId: extUserId}).exec(function(err, user) {
         if (user) {
-          FileClaimService.createNewClaim(user.id, callbacks);
+          FileClaimService.createNewClaim(user.id, callback);
```
ahsu1230: `callback` should just be a function instead of an object of functions.

winterfell/services/fileClaimService.js
```diff
@@ -8,39 +8,28 @@ function FileClaimService(app) {
...
-        } else {
-          _.isEmpty(callbacks) ? null : callbacks.onError(http.INTERNAL_SERVER_ERROR, httpErrors.DATABASE);
-          return null;
-        }
-      });
-    } else {
-      _.isEmpty(callbacks) ? null : callbacks.onError(http.BAD_REQUEST, httpErrors.CLAIM_INCOMPLETE_EXISTS);
+      return FileClaim.quickCreate(userId, callback);
+    }
+    if (_.isFunction(callback)) {
```
qwwqwwq: When would this argument not be a function? Or do you just want type-safe functions?

winterfell/models/document.js
```diff
@@ -8,8 +8,9 @@ var documentSchema = new Schema({
     key: String, // Code name for the form
     pdf: Buffer,
     user: Schema.Types.ObjectId
+}, {
+  timestamps: true
```
qwwqwwq: nice! super convenient

qwwqwwq:
One comment about the callbacks for creating new mongo documents

winterfell/services/fileClaimService.js
ahsu1230: how I've been doing callbacks is invoke a success if the expected object is defined (in this case `fileClaim` is not `undefined`) and invoke the error callback otherwise.

I had the `_.isFunction()` just to be more type-safe and to invoke the error callback, making the second argument `undefined`.

PR #36, opened 2016-08-12
User Address is now part of User object

ahsu1230:
* User address information is now part of a `contact` object inside `user`.
* Fix mailing service + tests


winterfell/test/testMailingService.js
```diff
@@ -3,10 +3,9 @@ var should = require('should');
 var fs = require('fs');
 
 var DestinationAddress = require('../models/destinationAddress');
-var UserAddress = require('../models/userAddress');
+var User = require('../models/user');
 var Letter = require('../models/letter');
 var Document = require('../models/document');
-var User = require('../models/document');
```
ahsu1230: typo

winterfell/test/testMailingService.js
```diff
@@ -56,29 +57,27 @@ describe('Mailing', function () {
...
-    var testUserAddressInstance;
     var testDestinationAddressInstance;
 
     before(function (done) {
         server = require('../app');
         service = new MailingService(server.app);
         service.Lob = mockLob;
 
-        User.create(testUser, function (userErr, user) {
+        User.remove({}, function() {
```
ahsu1230: I think before tests, it's good practice to reset the db as much as needed. This ensures other tests don't screw up what you're testing

winterfell/test/testMailingService.js
```diff
@@ -87,18 +86,18 @@ describe('Mailing', function () {
...
+            testUserInstance,
             testDestinationAddressInstance,
             [testDocumentInstance],
             function (sendLetterErr, letter) {
                 should.not.exist(sendLetterErr);
                 Letter.findOne({_id: letter._id}, function(queryErr, letter) {
                     should.not.exist(queryErr);
                     letter.user.should.deepEqual(testUserInstance._id);
-                    letter.sender.should.deepEqual(testUserAddressInstance._id);
+                    letter.sender.should.deepEqual(testUserInstance._id);
```
ahsu1230: what's the difference between `sender` and `user` in a `Letter`?

ahsu1230:
@qwwqwwq there are some mailing service things that this touches.

winterfell/test/testMailingService.js
qwwqwwq: I agree, eventually we should set it up so that then NODE_ENV=test we use a different copy of the database, vs when we run locally

winterfell/test/testMailingService.js
qwwqwwq: My idea was sender is the return address

qwwqwwq:
LGTM!

PR #39, opened 2016-08-13
add endpoint to create or reset test users

ahsu1230:
To better work on the front end, it would be much easier to log in as a test user who is guaranteed to start at a specific place.
This endpoint is meant to either create or reset test users back to their intended starting places.

As a starting point for creating these test users... they will only deal with FileClaims for now
* Ned will have no pending claims
* Robb has one pending claim
* Arya has submitted one claim
* Sansa has one processed claim


qwwqwwq:
This is pretty useful, LGTM

PR #40, opened 2016-08-14
Add Validation and Parsing to Forms

qwwqwwq:
- Upgrades JQuery / Lodash libraries
- Adds moment js angular wrapper
- Adds bootstrap CSS


qwwqwwq:
Adding bootstrap CSS kinda broke some of the other stylings, let me know if you think its fixable, if  its to hard we can remove and pursue something else. Bootstrap is really useful for form elements so I think it would be worth the effort if we can have it.

winterfell/webapp/src/js/app.js
```diff
@@ -4,54 +4,57 @@
...
     })
 
     .when("/form/:formId", {
-        templateUrl: "templates/form.html"
+      templateUrl: "templates/form.html"
     })
 
     // Otherwise
     .otherwise({
-        redirectTo: '/'
```
ahsu1230: woops thanks for all the indentation fixes

winterfell/webapp/src/js/formCtrl.js
```diff
@@ -20,10 +24,18 @@ app.controller("formCtrl", ['$scope', 'formRenderingService', 'formTemplateServi
...
+        $scope.model = {};
         $scope.fields = formTemplateService[$routeParams.formId];
 
-        $scope.model = {};
+        for (var i=0; i<$scope.fields.length; i++) {
+            if ($scope.fields[i].key.indexOf('date_signed') !== -1) {
+                $scope.model[$scope.fields[i].key] = currentDate();
+            }
+        }
     }]);
```
ahsu1230: a real small nit, but just fyi: let's try to end all files with a new line.
http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

ahsu1230:
lgtm

winterfell/webapp/src/js/formCtrl.js
qwwqwwq: agreed, will setup my ide to do this

PR #42, opened 2016-08-14
fix claim tests

ahsu1230:
some inconsistencies with some endpoints which caused some problems for one of the unit tests.

@qwwqwwq


qwwqwwq:
LGTM

PR #44, opened 2016-08-18
Adds Save Button To Form

qwwqwwq:
- Adds save button to form
- Add claim id to route
- Fix missing '/' in net.js

### DIFF
--- a/winterfell/controllers/adminController.js
+++ b/winterfell/controllers/adminController.js
@@ -112,7 +112,7 @@ module.exports = function (app) {
     });
   }
 
-  app.get('/admin/reset-test-users', function (req, res) {
+  app.post('/admin/reset-test-users', function (req, res) {
     console.log('Reset test users requested...');
     removeTestUsers(function() {
       _.forEach(allTestUsers, function(userToCreate) {
--- a/winterfell/controllers/adminController.js
+++ b/winterfell/controllers/adminController.js
@@ -112,7 +112,7 @@ module.exports = function (app) {
     });
   }
 
-  app.post('/admin/reset-test-users', function (req, res) {
+  app.get('/admin/reset-test-users', function (req, res) {
     console.log('Reset test users requested...');
     removeTestUsers(function() {
       _.forEach(allTestUsers, function(userToCreate) {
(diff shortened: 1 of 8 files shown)


winterfell/controllers/adminController.js
```diff
@@ -112,7 +112,7 @@ module.exports = function (app) {
     });
   }
 
-  app.post('/admin/reset-test-users', function (req, res) {
+  app.get('/admin/reset-test-users', function (req, res) {
```
ahsu1230: I know there's no body in this request so we can get away with a `GET`, but I think it's good practice to differentiate the notion between retrieving things vs. changing things.
http://stackoverflow.com/questions/504947/when-should-i-use-get-or-post-method-whats-the-difference-between-them

So since this endpoint is potentially changing users, let's keep it as `POST`

winterfell/webapp/src/js/app.js
```diff
@@ -49,7 +49,7 @@ app.config(function ($routeProvider) {
       templateUrl: "templates/claimView.html"
     })
 
-    .when("/form/:formId", {
+    .when("/form/:claimId/:formId", {
```
ahsu1230: small nit, but I think it's more RESTful if it is `claim/:claimId/form/:formId`
It's also more consistent with the rest of our endpoints that use claimIds.
Like `/claim/:claimId/dosomething`

winterfell/webapp/src/templates/form.jade
```diff
@@ -11,4 +11,6 @@ div(ng-controller="formCtrl")
     button(ng-click="signature = accept()")
         | Accept
     button(ng-click="render()")
-        | Download
\ No newline at end of file
+        | Download
+    button(ng=click="save()")
+        | Save
```
ahsu1230: nit: add extra line please after

ahsu1230:
some small nits, but if resolved, lgtm! 👍

winterfell/controllers/adminController.js
qwwqwwq: sure sounds good to me

winterfell/webapp/src/js/app.js
qwwqwwq: ah i see what you mean, given the resource is form under a claim, that makes sense to me that that would be the path

PR #45, opened 2016-08-19
Create Middleware to Verify Authentication

qwwqwwq:
- Middleware to check if user is authed, eventually we should use this to replace many one-off places in the code were we try to establish authorization
- Used this middleware within the fileClaim controller
- Updated & simplified fileClaimController tests

### DIFF
--- a/winterfell/middlewares/auth.js
+++ b/winterfell/middlewares/auth.js
@@ -1,0 +1,23 @@
+var http = require('http-status-codes');
+
+function Auth(app) {
+  this.app = app;
+}
+
+module.exports = Auth;
+
+module.exports.authenticatedOr404 = function (req, res, next) {
+  if (req.session.userId) {
+    return next();
+  }
+
+  console.log("Not authed");
+  res.sendStatus(http.NOT_FOUND);
+};
+
+module.exports.authenticatedOrRedirect = function (req, res, next) {
+  if (req.session.userId)
+    return next();
+
+  res.redirect('/');
+};
--- a/winterfell/controllers/claimController.js
+++ b/winterfell/controllers/claimController.js
@@ -56,7 +56,7 @@ module.exports = function (app) {
 
     User.findById(req.session.userId).exec(function (userErr, user) {
       if (user) {
-        FileClaimService.createNewClaim(user.id, callback);
+        ClaimService.createNewClaim(user.id, callback);
       } else {
         res.status(http.NOT_FOUND).send({error: httpErrors.USER_NOT_FOUND});
       }
@@ -96,7 +96,7 @@ module.exports = function (app) {
     console.log('[submitClaim] request received ' + JSON.stringify(req.body));
 
     handleClaimStateChange(req.params.extClaimId,
-      FileClaim.State.SUBMITTED,
+      Claim.State.SUBMITTED,
       claimUpdateCallbackFactory(res));
   });
 
@@ -104,7 +104,7 @@ module.exports = function (app) {
     console.log('[deleteClaim] request received for ' + req.params.extClaimId);
 
     handleClaimStateChange(req.params.extClaimId,
-      FileClaim.State.DISCARDED,
+      Claim.State.DISCARDED,
       claimUpdateCallbackFactory(res));
   });
 
--- a/winterfell/controllers/claimController.js
+++ b/winterfell/controllers/claimController.js
@@ -1,4 +1,3 @@
-var _ = require('lodash');
 var mongoose = require('mongoose');
 var http = require('http-status-codes');
 var httpErrors = require('./../utils/httpErrors');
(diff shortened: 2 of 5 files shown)


winterfell/controllers/fileClaimController.js
```diff
@@ -29,7 +30,7 @@ module.exports = function (app) {
   });
 
   // Get a particular claim
-  app.get('/claims/:extClaimId', function (req, res) {
+  app.get('/claim/:extClaimId', auth.authenticatedOr404, function (req, res) {
```
ahsu1230: should be `/claims/...`

ahsu1230: well what do you think? We can keep it singular `/claim/:claimId`, but we should be consistent across the board.

winterfell/controllers/fileClaimController.js
```diff
@@ -40,71 +41,68 @@ module.exports = function (app) {
     });
   });
 
-  app.post('/claims/create', function (req, res) {
+  app.post('/claims/create', auth.authenticatedOr404, function (req, res) {
     console.log('[createClaim] request received for ' + JSON.stringify(req.body));
-    var extUserId = req.body.extUserId;
     var callback = function (err, claim) {
       if (claim) {
-        res.status(http.OK).send({claim: FileClaim});
```
ahsu1230: my bad that was dumb

winterfell/controllers/fileClaimController.js
```diff
@@ -40,71 +41,68 @@ module.exports = function (app) {
...
+        res.status(http.CREATED).send(claim);
```
ahsu1230: in general, I like to always send an object as a response.
Consider when one day, we want to add more sibling objects to claim (let's call it `properties`)
So then we want to send()
```
{
  claim: claim,
  properties: props
}
```

We are more prone to backwards compatibility issues if we are coming from
```
.send(claim)
```
since the front-end will have to change. It gets even more hectic if multiple places on the front-end have to change for this. But then again, this is just a website (deploys instantly unlike apps and extensions) and we probably will only have one place to call this endpoint. Just my two cents from past experiences.

TLDR: Can we keep it as `.send({claim: claim})` please? 👻

winterfell/controllers/fileClaimController.js
```diff
@@ -40,71 +41,68 @@ module.exports = function (app) {
...
       }
-    } else {
-      res.status(http.BAD_REQUEST).send({error: httpErrors.INVALID_CLAIM_ID});
     }
+  }
+
+  app.post('/claim/:extClaimId/submit', function(req, res) {
+    console.log('[submitClaim] request received ' + JSON.stringify(req.body));
+
+    handleClaimStateChange(req.params.extClaimId,
```
ahsu1230: nice tyty

winterfell/middlewares/auth.js
```diff
@@ -0,0 +1,23 @@
...
+  console.log("Not authed");
+  res.sendStatus(http.NOT_FOUND);
+};
+
+module.exports.authenticatedOrRedirect = function (req, res, next) {
+  if (req.session.userId)
+    return next();
+
+  res.redirect('/');
+};
```
ahsu1230: nit: extra line at end of file please
Sorry I'm such a 👿 
http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

winterfell/test/testFileClaimController.js
```diff
@@ -42,37 +42,31 @@ describe('FileClaimController', function() {
...
+      .get('/claims')
+      .expect(http.NOT_FOUND, done);
   });
 
-  it('Create claim for user - user dne', function(done) {
-    testSession
-      .post('/claims/create')
-      .send({ extUserId: 'asdf' })
-      .expect(http.NOT_FOUND, done);
+  it('should sign in', function (done) {
```
ahsu1230: this test should not be in fileclaimcontroller test

winterfell/middlewares/auth.js
```diff
@@ -0,0 +1,23 @@
+var http = require('http-status-codes');
+
+function Auth(app) {
+  this.app = app;
+}
+
+module.exports = Auth;
+
+module.exports.authenticatedOr404 = function (req, res, next) {
```
ahsu1230: this is awesome

winterfell/controllers/fileClaimController.js
```diff
@@ -40,71 +41,68 @@ module.exports = function (app) {
...
       } else {
-        res.sendStatus(statusCode);
+        res.sendStatus(http.NOT_FOUND);
       }
-    } else {
-      res.status(http.BAD_REQUEST).send({error: httpErrors.INVALID_CLAIM_ID});
     }
+  }
+
+  app.post('/claim/:extClaimId/submit', function(req, res) {
```
ahsu1230: `/claim` or `/claims`?

winterfell/test/testFileClaimController.js
```diff
@@ -42,37 +42,31 @@ describe('FileClaimController', function() {
...
+    testSession.post('/auth/login')
+      .send({email: targetUser.email, password: targetUser.password})
+      .expect(200, done);
   });
 
-  it('Create claim for user - success', function(done) {
+
+  it('Create claim for user - user dne', function(done) {
```
ahsu1230: misleading test name
should be -success case

ahsu1230:
couple of comments, but mostly good.
Let's make sure to decide on `claim/` vs. `claims/` so the front end can change appropriately as well.

winterfell/controllers/fileClaimController.js
qwwqwwq: the RESTful convention I was aiming for was: use plural for resources that return an array of that model, and singular for resources that return a single instance of the model, so since this is returning just one claim I wanted to change to singular

winterfell/controllers/fileClaimController.js
qwwqwwq: Ok to the convention is we return an object which is of the form:

`{ <model_name>: <model_object> }`

that sounds reasonable to me

winterfell/middlewares/auth.js
qwwqwwq: thanks, i agree, sorry I just need to setup my IDE to do this

winterfell/test/testFileClaimController.js
qwwqwwq: oh this is just to establish an authorized session for the other tests

PR #46, opened 2016-08-19
Feature idme

ahsu1230:
still a draft... should not merge until clientId for idme is in biscuit

The authorization with idme also won't work until we get authorized by them to grant us permissions to request the `military` scope of users


winterfell/controllers/authController.js
```diff
@@ -1,5 +1,6 @@
 var _ = require('lodash');
 var passport = require('passport');
+var requestify = require('requestify');
```
qwwqwwq: Do we need a new library just to do this POST request to ID.me? Would prefer to minimize the number of external libs we have, it will make the project a lot easier to maintain

winterfell/webapp/src/js/net.js
```diff
@@ -1,12 +1,26 @@
...
+  if (dev) {
+    baseUrl = "http://localhost:3999";
+  } else {
+    baseUrl = "https://www.vetafi.org";
+  }
+
+  // social properties
+  var socialIdMe = {
+    clientId: '71ffbd3f04241a56e63fa6a960fbb15e',
+    redirectServerUri: baseUrl + "/auth/link/idme",
```
qwwqwwq: You don't need to include the base url (https://www.vetafi.org) in the request,  the AJAX default is to resolve the path relative to the current page.. so this could be just `redirectServerUri: "/auth/link/idme"` and the above logic about what the base url is in dev vs prod can be safely removed

winterfell/controllers/authController.js
ahsu1230: Mentioned this elsewhere:
I used requestify for now because I was getting access-control-allow-origin problems from using the jquery post methods (ajax post do not work when the domain of your current site does not match the domain of your request). Requestify at least helps with all that complications - we should look for a workaround soon 👍

ahsu1230:
Sorry I'm going to merge this for now to clean up our PR lists. We can talk about requestify sometime soon

ahsu1230:
i'll also note that Id.Me still hasn't responded back to me related to being able to access military scope. I'll give them a phone call soon to talk with a representative

PR #47, opened 2016-08-19
Ops gulp create xhrenv

ahsu1230:
we can use gulp to create a dev webapp or a prod webapp depending on how we call gulp.

For example, by default, will build a debug webapp. Base server url is also pointing to localhost. etc.
```
$ gulp
```

```
$ gulp --release
```
will build the webapp with javascript minified, base server pointing to our real one. etc.


ahsu1230:
This PR doesn't necessarily need to depend on this other PR:
https://github.com/bayeshack2016/vetafi/pull/46

but it made more sense to make this change after looking at some shortcomings of that other PR

qwwqwwq:
Nice I like the approach, LGTM

qwwqwwq:
Still think we dont need requestify though

ahsu1230:
I used requestify for now because I was getting `access-control-allow-origin` problems from using the jquery `post` methods (ajax post do not work when the domain of your current site does not match the domain of your request). Requestify at least helps with all that complications - we should look for a workaround soon 👍

PR #51, opened 2016-08-25
Footer directive (simple example of using bootstrap)

ahsu1230:
Few things:
* Created a new gulp tasks that takes all of our third-party `css` files from libraries to be condensed into its own `libs.css` file. 
* We want to make sure we import `libs.css` before our `main.css` so that any of our classes should ideally override what is in `libs.css`. This organization also helps us to figure out what properties are coming from which class from which file.
* Created a directive that handles the footer. It just has links, and image and a disclaimer message. It uses bootstrap to help us with responsiveness. 

The main focus here is to agree with what's a good way for a template to have both bootstrap and our own class properties.


winterfell/webapp/src/styles/standards.styl
```diff
@@ -31,6 +31,10 @@ h5_size = 15px							// 17px
 h5_lineHeight = 1.3em				// 22px
 h5_letterSpacing = 1px			// 1px
 
+h6_size = 13px							// 13px
```
ahsu1230: not sure why this spacing is so bad

qwwqwwq:
LGTM i like having the library css in another file

PR #53, opened 2016-08-29
Fix Client Side Requests

qwwqwwq:
- Don't use the base url for XHR requests, its not required, we only need this information for OAuth protocol
- Fix the paths in net.js


ahsu1230:
cool, thanks for fixing the paths. lgtm

PR #54, opened 2016-08-29
Wire Together Claim Flow

qwwqwwq:
- Confirmed flow for starting and creating claim works.
- Moved TOS to a modal so that it does not interrupt flows.

### DIFF
--- a/winterfell/services/claimService.js
+++ b/winterfell/services/claimService.js
@@ -5,19 +5,11 @@ var Claim = require('./../models/claim');
 
 function ClaimService(app) {
   this.app = app;
-};
+}
 
 module.exports = ClaimService;
 module.exports.createNewClaim = function(userId, callback) {
-  return Claim.findOne({ userId: userId, state: Claim.State.INCOMPLETE }).exec(function(err, fileClaim) {
-    if (_.isEmpty(fileClaim)) {
-      return Claim.quickCreate(userId, callback);
-    }
-    if (_.isFunction(callback)) {
-      callback();
-    }
-    return null;
-  });
+  return Claim.quickCreate(userId, callback);
 };
 
 module.exports.addForm = function(claimId, file, callbacks) {
--- a/winterfell/services/claimService.js
+++ b/winterfell/services/claimService.js
@@ -8,8 +8,16 @@ function ClaimService(app) {
 }
 
 module.exports = ClaimService;
-module.exports.createNewClaim = function(userId, callback) {
-  return Claim.quickCreate(userId, callback);
+module.exports.findIncompleteClaimOrCreate = function(userId, callback) {
+  return Claim.findOne({ userId: userId, state: Claim.State.INCOMPLETE }).exec(function(err, fileClaim) {
+    if (_.isEmpty(fileClaim)) {
+      return Claim.quickCreate(userId, callback);
+    }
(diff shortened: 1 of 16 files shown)


winterfell/services/claimService.js
```diff
@@ -5,19 +5,11 @@ var Claim = require('./../models/claim');
 
 function ClaimService(app) {
   this.app = app;
-};
+}
 
 module.exports = ClaimService;
 module.exports.createNewClaim = function(userId, callback) {
-  return Claim.findOne({ userId: userId, state: Claim.State.INCOMPLETE }).exec(function(err, fileClaim) {
```
qwwqwwq: I think create new claim should always do what it says, create a new claim. Checking for an existing claim and that sort of logic should be at a higher level.

winterfell/webapp/src/js/tosCtrl.js
```diff
@@ -2,15 +2,12 @@ var app = angular.module('vetafiApp');
 app.controller('tosCtrl', ['$scope', '$location', 'claimService',
   function($scope, $location, claimService) {
     $scope.onAccept = function() {
-      var input = $('.vfi-checkbox input');
-      if (input.prop('checked')) {
+      console.log("onAccept", $scope);
```
ahsu1230: extra console.log

winterfell/webapp/src/js/claimStartCtrl.js
```diff
@@ -1,18 +1,22 @@
...
+    }
+
+    $scope.onClickNext = function () {
+      if (!claimService.acceptedTos()) {
+        if (!getTosAgreement()) {
+          return;
         }
+      }
+      net.startClaim().then(function (res) {
+        console.log('/claim/' + res.data.claim.externalId + '/select-forms');
```
ahsu1230: console.log

winterfell/webapp/src/js/claimSelectFormsCtrl.js
```diff
@@ -1,33 +1,11 @@
...
-        description: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor.',
-        category: '',
-        required: false,
-        completed: true
-      }
-    ];
+    $scope.claimId = $routeParams.claimId;
+    $scope.forms = formTemplateService;
+
+    console.log(formTemplateService);
```
ahsu1230: still need this?

winterfell/services/claimService.js
ahsu1230: I guess it's a bit weird that now this is reduced to one simple statement.
I don't think we should ever have more than one INCOMPLETE claim. `Claim.quickCreate()` is the part you described that just creates a new claim. However, if multiple controllers / services want to create a claim, they should have the logic to check existing incomplete claims. Rather than having this logic in all those other places, I think it should be consolidated here.

ahsu1230:
mostly lgtm other than one major reply to your comment.

winterfell/services/claimService.js
qwwqwwq: Hmm ok I have an idea, maybe we can rename this method createOrFindExistingClaim and then have the `new claim` button read `finish in progress claim` if the user has a started claim.

PR #55, opened 2016-08-29
Feature idme sign up (create new user)

ahsu1230:
This PR has the following changes:

* First off, we now have two Vetafi Applications with ID.Me. One is for prod, and the other is for a local dev sandbox. As such, when we build our app, the correct Id.Me clientId needs to correspond to the prod or dev IdMe application.
* The xhrEnv is more useful now because it contains `clientId` and `redirectUri` which depends on whether we build in dev or release.
* xhrEnv is its own gulptask which will generate a `xhrEnv.js` depending on dev or release. The previous PR related to this actually added both dev and release into the build if you ran it twice (once dev and once release). Manually including the `xhrEnv.js` forces us to always pick only one.

Here's an explanation of how Id.Me works.
1) When a user clicks a Id.Me button, we redirect them to: `https://api.id.me/oauth/authorize...` Here, they will sign in to their Id.Me account and give Vetafi access to their Id.Me info.
2) After signing in, Id.Me will redirect to our server endpoint: `/auth/link/idme/`. Here, we get this code which we will use to obtain an access token.
3)  We POST query for an access token to here: `https://api.id.me/oauth/token`. If an access token is not acquired by 5 minutes, the code from step2 will expire.
4) Then, once an access token is acquired, we will use it to get some user info. Use a GET request to `https://api.id.me/api/public/v2/attributes.json?access_token=ACCESS_TOKEN`. 
The info we get is the following:
```
{
  "id": "65593b840f5c4b1391fb862561efceab",
  "verified": true,
  "affiliation": "Service Member",
  "fname": null,
  "lname": null,
  "zip": null,
  "uuid": "65593b840f5c4b1391fb862561efceab",
  "email": "partner+vetafi_sm_1@id.me"
}
```
Look here for more info: https://developer.id.me/documentation

4) From here, if everything is fine and dandy, we will create a new user with the given email, firstname, lastname. Make sure that the social user is added to the user.
5) SocialUser is now a subschema (not its own table). A user will have an array of SocialUsers.

For now, this change only deals with creating a new user through Id.Me.
Linking an Id.Me account to an existing user will come in a different change.

Also, the test user to use is this:
```
partner+vetafi_sm_1@id.me
2eR7OG5DAcB3
```


winterfell/controllers/authController.js
```diff
@@ -92,25 +93,96 @@ module.exports = function (app) {
...
-  // Endpoint for routing idme authorization
+  // Server endpoint for redirecting from ID.me authorization
   // https://api.id.me/oauth/authorize?...
   app.get('/auth/link/idme', function(req, res) {
-    console.log('[authLinkIdMe] link request with ' + JSON.stringify(req.params));
-    var code = req.params.code;
+    var prodEnv = app.environment == constants.environment.PROD;
+    var baseUri = prodEnv ? 'https://www.vetafi.org' : 'http://localhost:3999';
+    var clientId = prodEnv ? '71ffbd3f04241a56e63fa6a960fbb15e' : '684c7204feed7758b25527eae2d66e28';
+    var secretClientId = prodEnv ? 'some-secret' : '57ebec28ad4bae403d0a2702f2f81801';
```
qwwqwwq: Wait is this our actual ID.me secret?

ahsu1230: I didn't put the real prod secretClientId. That other one is our development secret client id. You can regenerate secret client ids, so I'll go ahead and do that once both of those ids are in biscuit.

qwwqwwq: ok cool, we can release the above 3 lines with calls to `biscuitService` after my pull is landed

winterfell/controllers/authController.js
```diff
@@ -92,25 +93,96 @@ module.exports = function (app) {
...
+      };
+      console.log('[authLinkIdMe] request token with data: ' + JSON.stringify(data));
+      request.post({
+        url: 'https://api.id.me/oauth/token',
+        json: true,
+        body: data
+      }, function(error, response, body) {
+        var accessToken = body.access_token; // provided by ID.me
+        if (accessToken) {
+          request.get('https://api.id.me/api/public/v2/attributes.json?access_token=' + accessToken, function(accessError, accessResponse, accessBody) {
```
qwwqwwq: lets move the `api.id.me` url to constants or configuration file

winterfell/services/userService.js
```diff
@@ -77,3 +79,74 @@ module.exports.upsertUserValues = function(userId, keyValues, callback) {
...
+ *  email
+ * }
+ */
+module.exports.createNewUserFromIdMe = function(idmeInfo, callback) {
+  var newUser = {
+    firstname: idmeInfo.fname,
+    middlename: null,
+    lastname: idmeInfo.lname,
+    email: idmeInfo.email,
+    password: "fakepassword", // todo: generate a short random 8-char password
```
qwwqwwq: Probably they should just have no password if they are using oAuth right? They will authenticate with oAuth and not password oAuth from then on..

winterfell/webapp/src/js/profileCtrl.js
```diff
@@ -68,9 +68,8 @@ app.controller('profileCtrl', ['$scope', '$location', 'profileService', 'claimSe
     };
 
     $scope.clickLinkIdMe = function() {
-      net.linkIdMe().then(function(resp) {
-        debugger;
-      });
+      // Redirect to IdMe auth page, which will redirect to the specified uri setup with IdMe.
+      $window.location.href = net.getAuthIdMeUrl();
```
qwwqwwq: Eventually we might want to do this in a modal or iframe or something

winterfell/webapp/src/js/xhrEnv/xhrEnvRelease.js
```diff
@@ -4,5 +4,7 @@ app.factory('xhrEnv', function() {
   return {
     isDev: false,
     baseServerUrl: 'https://www.vetafi.org',
+
+    idMeClientId: '71ffbd3f04241a56e63fa6a960fbb15e',
```
qwwqwwq: remove after https://github.com/bayeshack2016/vetafi/pull/57 is landed

winterfell/webapp/src/js/xhrEnv/xhrEnvDev.js
```diff
@@ -4,5 +4,7 @@ app.factory('xhrEnv', function() {
   return {
     isDev: true,
     baseServerUrl: 'http://localhost:3999',
+
+    idMeClientId: '684c7204feed7758b25527eae2d66e28'
```
qwwqwwq: remove after https://github.com/bayeshack2016/vetafi/pull/57 is landed

winterfell/controllers/authController.js
```diff
@@ -92,25 +93,96 @@ module.exports = function (app) {
     });
   });
 
-  // Endpoint for routing idme authorization
+  // Server endpoint for redirecting from ID.me authorization
   // https://api.id.me/oauth/authorize?...
   app.get('/auth/link/idme', function(req, res) {
```
qwwqwwq: Cant we just use passport.js for all this? They have a tidy implementation for handling the post redirect, something like:

```
var passport = require('passport')
  , OAuth2Strategy = require('passport-oauth').OAuth2Strategy;

passport.use('provider', new OAuth2Strategy({
    authorizationURL: 'https://www.provider.com/oauth2/authorize',
    tokenURL: 'https://www.provider.com/oauth2/token',
    clientID: '123-456-789',
    clientSecret: 'shhh-its-a-secret'
    callbackURL: 'https://www.example.com/auth/provider/callback'
  },
  function(accessToken, refreshToken, profile, done) {
    User.findOrCreate(..., function(err, user) {
      done(err, user);
    });
  }
));
```

See http://passportjs.org/docs/oauth

Also I noticed from the comment you imply this controller is just for authorization. I think there only needs to be one controller for both authorization and login with OAuth2.0, its all handled on the ID.me side based on what our client id is and whether the user has authorized that client id before.

winterfell/services/userService.js
ahsu1230: we will need a way to prevent them from entering email + password then.
one way could be to have a boolean flag that prevents them to. Or generate a random password that we never give to the user and it won't work if they try. When we generate a random password, if a user ever wants to go through email, they could always reset it.

config/biscuit/test_secrets.yaml
```diff
@@ -1,5 +1,11 @@
 _keys:
 - algorithm: none
+test::id-me-client-id:
```
ahsu1230: tests don't pass without these

winterfell/controllers/authController.js
```diff
@@ -1,6 +1,7 @@
 var _ = require('lodash');
 var passport = require('passport');
-var requestify = require('requestify');
+var request = require('request');
```
ahsu1230: unused here

winterfell/app.js
```diff
@@ -4,7 +4,8 @@ var session = require('express-session');
 var fs = require('fs');
 var path = require('path');
 var bodyParser = require('body-parser');
-var ENVIRONMENT = require('./utils/constants').environment;
+var constants = require('./utils/constants');
```
ahsu1230: Constants (capital C)

winterfell/app.js
```diff
@@ -38,14 +40,20 @@ app.get('/', function(req, resp) {
 // Connect to a mongodb server using mongoose
 require('./config/mongoose')(environment);
 
-// Setup lob api
-biscuit.get(environment + '::lob-api-key', function(err, secret) {
-  if (err) {
-    throw err;
-  }
+// Setup biscuit keys
+function setupBiscuitKey(keyName) {
```
qwwqwwq: i like it

winterfell/middlewares/passport.js
```diff
@@ -48,4 +55,25 @@ module.exports = function (app) {
             }
         });
     }
+
+    passport.use('idme', new OAuth2Strategy({
+        authorizationURL: 'https://api.id.me/oauth/authorize',
+        tokenURL: 'https://api.id.me/oauth/token',
+        clientID: idmeClientId,
+        clientSecret: idmeSecretId,
+        callbackURL: '/auth/idme/callback'
```
qwwqwwq: Not 100% sure but I think this needs to be a full url, i.e. https://www.vetafi.org etcc

winterfell/services/userService.js
qwwqwwq: Hmm, why cant we just leave the password null? I think they could definitely make a username/password if they want, in addition to oAuth, but I guess having a static default value for password seems a little dangerous, because if we mess up our logic, it would give people access to any oAuth account

winterfell/services/userService.js
```diff
@@ -77,3 +79,74 @@ module.exports.upsertUserValues = function(userId, keyValues, callback) {
...
+  });
+};
+
+// Method to add an ACTIVE socialUser to a User
+module.exports.addSocialUser = function(userId, type, token, callback) {
+  var now = Date.now();
+  User.findOne({_id: userId}, function(err, user) {
+    user.socialUsers.push({
+      type: type,
+      oauthToken: token,
```
qwwqwwq: are we supposed to save this token in a database?

qwwqwwq:
LGTM! had a few questions comment but I think we should go ahead and land

winterfell/services/userService.js
ahsu1230: I'm pretty sure. Normally you use that token to make future queries with social providers. I haven't fully looked at what you could with Id.me but I know it's a thing for facebook.

winterfell/middlewares/passport.js
ahsu1230: hmm it worked when I tried it locally. But I think you're right, it should be a full url.

PR #56, opened 2016-09-01
Redo home page, using bootstrap grid

ahsu1230:
Redo home page to have bootstrap grid system and responsiveness.

Works pretty well!
Still figuring things out when it comes to good practice especially when it comes to working with bootstrap, but I think I have several good ideas:

* `vfi-` prefix is unnecessary for classes. It's too verbose. However, to better avoid conflicting properties between classes, every view should have an id. For example, `vfi-home-view` or `vfi-footer-view`, then make sure in the styl, every class first searches within this id.
```
#vfi-home-view
  & .banner
     ...
```
turns to
```
#vfi-home-view .banner {
...
}
```
Just having `.banner` is very prone to conflicting properties.

* our main.styl should have `!important` on all properties so bootstrap doesn't override properties. We were missing font styles for the longest time because bootstrap was overriding everything even though I included `boostrap.css` before our `main.css`


qwwqwwq:
Nice looking really good! Is this still in progress?

Also I would be careful about overusing `!important`, if you need to use that it usually indicates other problems that will come back to haunt us. If you want we can compile bootstrap differently if you want to change the base styles.

Also for using all CSS classes relative to parent id, don't you think this will make our stylesheets less DRY? We would need to repeat classes for similar elements in different views right? Or does .styl have ways to accommodate this?

ahsu1230:
well as of now, we don't really have any views that have the same stuff, so I think it's better to separate everything until we need to combine elements together (in which case, should be using a directive with its own template). Most main elements like buttons, links, text, headers, etc. are all made in `main.styl` and apply throughout the entire site, so you wouldn't have to repeat code for those.
Does that answer the question?

PR #58, opened 2016-09-05
Improve Authentication Handling in Frontent

qwwqwwq:
- Switch to UI router
- Setup hierarchical components with all components descending from the header controller
- Force resolution of user when home is loaded
- Fix `/user` endpoint to read session info
- Use cookie expiration instead of redis session expiration.
- Move inactivity handler logic server side by pinging touch on interval


winterfell/controllers/userController.js
```diff
@@ -4,13 +4,14 @@ var httpErrors = require('./../utils/httpErrors');
...
 var UserValues = require('../models/userValues');
 var UserService = require('./../services/userService');
+var auth = require('../middlewares/auth');
 
 module.exports = function (app) {
 
   // Get a user's information based on externalId
-  app.get('/user/:extUserId', function (req, res) {
+  app.get('/user', auth.authenticatedOr404, function (req, res) {
     console.log('[getUser] request received for ' + req.params.extUserId);
```
ahsu1230: replace with req.session.userId?

winterfell/webapp/src/js/app.js
```diff
@@ -4,56 +4,126 @@
...
+      }]
+    }
+  });
+
+  $urlRouterProvider.otherwise('/app');
+
+  // Home
+  $stateProvider.state({
+    name: 'root.home',
+    url: '',
```
ahsu1230: does url need to be `/`?

winterfell/webapp/src/js/app.js
```diff
@@ -4,56 +4,126 @@
...
-    .when("/claim/:claimId/form/:formId", {
-      templateUrl: "templates/form.html"
-    })
-
-    // Otherwise
-    .otherwise({
-      redirectTo: '/'
-    });
-});
+app.config(['$stateProvider', '$urlRouterProvider', function ($stateProvider, $urlRouterProvider) {
```
ahsu1230: nice this is cool

winterfell/webapp/src/js/claimConfirmCtrl.js
```diff
@@ -1,23 +1,21 @@
...
 
     $scope.onClickConfirm = function() {
       var claim = claimService.getIncompleteClaim();
       if (!_.isEmpty(claim)) {
-        net.submitClaim(claim.id).then(function(resp) {
+        net.submitClaim($stateParams.claimId).then(function(resp) {
           // todo: set claim state or re-fetch all user claims?
           if (resp) {
-            $location.path("/claim/submitted");
+            $state.transitionTo("root.claimsubmit", {claimId: $stateParams.claimId});
```
ahsu1230: oh neat

winterfell/webapp/src/js/claimConfirmCtrl.js
```diff
@@ -1,23 +1,21 @@
...
           } else {
-            $location.path("/claim/submitted").search({error: 'missing'});
```
ahsu1230: these are in place because the template here depends on a `?error=missing` after the url.
We should probably have a better way to do this. Maybe if we're going the stateRouting way, we should have two views. One submitted success and one submitted error. There may also be different kinds of errors so the submitted error page should handle all of those.

winterfell/webapp/src/js/claimSelectFormsCtrl.js
```diff
@@ -1,14 +1,9 @@
...
-  function($scope, $location, claimService, formTemplateService, $routeParams) {
-
-    $scope.claimId = $routeParams.claimId;
+app.controller('claimSelectFormsCtrl', ['$scope', 'claimService', 'formTemplateService', '$stateParams',
+  function($scope, claimService, formTemplateService, $stateParams) {
+    $scope.claimId = $stateParams.claimId;
     $scope.forms = formTemplateService;
 
-    $scope.onClickFinish = function() {
-      $location.path('/claim/confirm');
```
ahsu1230: transition to `/claim/confirm`

ahsu1230:
```
$state.transitionTo('root.claimselect', {claimId:$stateParams.claimId});
```

winterfell/webapp/src/js/app.js
```diff
@@ -4,56 +4,126 @@
...
+  $stateProvider.state({
+    name: 'root.profile',
+    url: '/profile',
+    templateUrl: 'templates/profile.html',
+    controller: 'profileCtrl',
+    abstract: true
+  });
+
+  $stateProvider.state({
+    name: 'root.profile.military',
```
ahsu1230: thanks for keeping these subcategories

winterfell/webapp/src/js/claimViewCtrl.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 var app = angular.module('vetafiApp');
-app.controller('claimViewCtrl', ['$scope', '$location', 'claimService', 'net', 'ngDialog',
-  function($scope, $location, claimService, net) {
-    $scope.claimId = $location.path().substring(7);
```
ahsu1230: lol I'm embarrassed this hack got found ^^

winterfell/webapp/src/js/profileService.js
```diff
@@ -1,12 +1,47 @@
 'use strict';
 var app = angular.module('vetafiApp');
-app.factory('profileService', ['claimService', function(claimService) {
-  return {
-    userInfo: {},
-    clearUserInfo: function() {
-      sessionStorageHelper.removePair(vfiConstants.keyUserId);
-      this.userInfo = {};
+app.service('Profile', ['claimService', '$q', 'net', function(claimService, $q, net) {
```
ahsu1230: we have a Profile and profileService.
What do you think about renaming Profile to Account?

winterfell/webapp/src/js/profileService.js
```diff
@@ -1,12 +1,47 @@
...
+
+    this.setUser = function(user) {
+      this.user = user;
+    };
+
+    this.unsetUser = function() {
+      this.user = null;
+    };
+
+    this.isSetUser = function() {
```
ahsu1230: hasUser?

winterfell/webapp/src/templates/profile/military.jade
```diff
@@ -0,0 +1,17 @@
+div.vfi-profile-section#military
```
ahsu1230: thanks for moving things

winterfell/webapp/src/templates/profile/claims.jade
```diff
@@ -0,0 +1,18 @@
+div.vfi-profile-section#claims
+    div.vfi-section-content(ng-show="claims.length == 0")
+        | You have no pending file claims at the moment.
+        br
+        | Click <a href="#/claim/start">here</a> to start a file claim.
+    div.vfi-section-content(ng-class="{'vfi-incomplete': claim.state=='incomplete', 'vfi-submitted': claim.state=='submitted', 'vfi-processed': claim.state=='processed'}", ng-repeat="claim in claims", ng-show="claims.length > 0")
+        div.vfi-section-head
+            h4.vfi-date {{claim.dateMessage}}
+            div.vfi-action-wrapper
+                button.vfi-edit-btn(ng-click="root.claimselect({claimId: claim.id})", ng-if="claim.state=='incomplete'")
```
ahsu1230: this should be using `ui-sref` right?

winterfell/webapp/src/templates/profile/claims.jade
```diff
@@ -0,0 +1,18 @@
...
+                    div.vfi-icon.vfi-edit
+                button.vfi-discard-btn(ng-click="clickClaimDiscard(claim.id)", ng-if="claim.state=='incomplete'")
+                    div.vfi-icon.vfi-delete
+                button.vfi-view-btn(ng-click="root.claimview({claimId: claim.id})")
```
ahsu1230: should be using `ui-sref` right?

winterfell/middlewares/redis-session.js
```diff
@@ -22,8 +22,8 @@ module.exports = function (app) {
       store: new redisStore({
         host: host,
         port: port,
-        client: redisClient,
-        ttl: Constants.SESSION_EXPIRE_TIME}), // expires after 20 minutes
+        client: redisClient}), 
+      cookie: {expires: new Date(Date.now() + Constants.SESSION_EXPIRE_TIME)}, // expires after 20 minutes
```
ahsu1230: this will just auto set cookies for us on clientside?

winterfell/webapp/src/js/app.js
```diff
@@ -4,56 +4,126 @@
...
-    // Otherwise
-    .otherwise({
-      redirectTo: '/'
-    });
-});
+app.config(['$stateProvider', '$urlRouterProvider', function ($stateProvider, $urlRouterProvider) {
+
+  $stateProvider.state({
+    name: 'root',
+    url: '/app',
```
ahsu1230: can you explain why this needs a url `/app`?

winterfell/webapp/src/js/applicationCtrl.js
```diff
@@ -0,0 +1,8 @@
+var app = angular.module('vetafiApp');
+app.controller('ApplicationController', function ($scope) {
```
ahsu1230: is this used?

winterfell/webapp/src/js/claimSelectFormsCtrl.js
ahsu1230: oh i see the `ui-sref` nvm!

winterfell/webapp/src/js/profileService.js
```diff
@@ -1,12 +1,47 @@
...
-      sessionStorageHelper.removePair(vfiConstants.keyUserId);
-      this.userInfo = {};
+app.service('Profile', ['claimService', '$q', 'net', function(claimService, $q, net) {
+    this.user = null;
+
+    this.logout = function() {
       claimService.clearClaims();
-    }
-  };
+      this.unsetUser();
```
ahsu1230: do we not need to clear the cookies or anything if user logs out?

ahsu1230:
This is really great stuff! I don't want to bog you down with details since this is already a big change and a lot of other stuff will change soon anyway.

winterfell/webapp/src/js/app.js
qwwqwwq: Yeah so basically this is the "parent" state, and it will be at the root of our app, and all pages under it will have certain things in common (in this case, all pages under this will try to resolve the user)

We dont have to call it `/app`, we could just have it be `/`, but I thought it would be useful if we want to have another angular page tree, its alot easier to distinguish but giving this a real path

winterfell/webapp/src/js/app.js
qwwqwwq: By leaving the url empty here, `root.home` becomes the default state for the path `/app`

winterfell/webapp/src/js/claimViewCtrl.js
qwwqwwq: :P

winterfell/webapp/src/js/profileService.js
qwwqwwq: Redirecting to the logout page is what actually logs hte user our, because that is when the server tells the browser to unset the cookie. Because the express-session cookie is http only, we cant touch it from javascript, even to test if it exists. This just clears the service state for good measure

winterfell/webapp/src/js/profileService.js
qwwqwwq: I feel like `has` is for collection-like objects, `isSet` to me makes more sense for instance - fields

winterfell/webapp/src/templates/profile/claims.jade
qwwqwwq: oh yeah, never made it to this page my bad

winterfell/middlewares/redis-session.js
qwwqwwq: Yes exactly when the user hits the login link, express-session tells the browser to set an http only cookie. The http only cookie cannot be accessed in any way by the client side javascript, which makes it more robust to XSS attacks. The server will also set this expiration date on the cookie. When the server sees the cookie is older than the expiration date, the user will be automatically logged out. All of the real authentication stuff happens on the server side.

winterfell/webapp/src/js/applicationCtrl.js
qwwqwwq: nope, ended up using the header ctrl as the root controller

winterfell/webapp/src/js/claimConfirmCtrl.js
qwwqwwq: whoops meant to come back and fix this but forgot to, will fix

winterfell/webapp/src/js/profileService.js
qwwqwwq: I'm not sure what you mean, this is the only profileService, i replaced the old one, and changed it to be `app.service` so that it is a singleton

ahsu1230: oh my bad. I guess I could've sworn I saw something depend on both profileService and Profile. Nevermind then.

qwwqwwq:
Yeah looks like there might be some merge conflicts already, will take me a bit to sort out

PR #59, opened 2016-09-07
New login/signup (looks a lot better!)

ahsu1230:
still need to clean things up


winterfell/webapp/src/styles/signin.styl
```diff
@@ -1,97 +1,216 @@
 @import "standards.styl"
 
+backgroundUrl = '../icons/veteran-dog-bg.jpg'
```
qwwqwwq: I noticed this image is ~ 3 MB, makes the page slow even serving locally, not sure it'll be suitable for production, is there anything else we can use instead?

ahsu1230: ah yea. I was using a super big one. I was able to create a smaller scaled one that is 193KB. I may start purchasing a few photos that are licensed which we can legally use. I found a pretty good site that has really good veteran-related photos. Might start using those.

qwwqwwq:
Ok just checking, sounds good, LGTM!

PR #61, opened 2016-09-14
Fix errors on startup

ahsu1230:
@qwwqwwq


winterfell/webapp/src/js/profileService.js
```diff
@@ -34,8 +34,9 @@ app.service('Profile', ['claimService', '$q', 'net', function(claimService, $q,
 
       var that = this;
       net.getUserInfo().then(function success(res) {
-        that.setUser(res.data);
```
ahsu1230: these cause some jQuery errors: res may be undefined

ahsu1230:
ah crap. Accidentally merged changes together.
Look at commit 2e4589f

qwwqwwq:
LGTM

winterfell/webapp/src/js/profileService.js
qwwqwwq: It doesnt make sense to me how res would ever be undefined here..

qwwqwwq: Ah I see this is because of the "dev mode" XHR for httpGet. Is that still necessary? When do you have a need to test things without a server backend?

ahsu1230: hmm maybe not anymore. It made sense a while ago when I wanted to test how things flowed.

ahsu1230:
fixed now

PR #62, opened 2016-09-15
Ahsu redo confirm claim

ahsu1230:


winterfell/webapp/src/js/claimConfirmCtrl.js
```diff
@@ -1,8 +1,98 @@
...
+        name: 'Me',
+        email: 'ahsu@gmail.com'
+      },
+      {
+        name: 'Mom',
+        email: 'mom@gmail.com'
+      }
+    ];
+
+    $scope.formsList = [
```
qwwqwwq: Can you just inject `formTemplateService` instead? It has the list of forms

winterfell/webapp/src/js/claimConfirmCtrl.js
```diff
@@ -1,8 +1,98 @@
...
+app.controller('claimConfirmCtrl', ['$scope', '$state', 'net', 'claimService', '$stateParams', '$uibModal',
+  function($scope, $location, net, claimService, $stateParams, $uibModal) {
     $scope.claimId = $stateParams.claimId;
 
+    $scope.user = {
+      firstName: 'Aaron',
+      lastName: 'Hsu'
+    };
+
+    $scope.vaAddress = {
```
qwwqwwq: Shouldnt we get these and the following variables from the User model or some other model?

winterfell/webapp/src/js/claimConfirmCtrl.js
```diff
@@ -1,8 +1,98 @@
...
+      return openModifyContactModal('email', $scope.emailList, index);
+    };
+
+    function openModifyContactModal(forType, list, index) {
+      $scope.modalData = {
+        forType: forType,
+        list: list,
+        targetIndex: index >= 0 ? index : -1
+      };
+      return $uibModal.open({
```
qwwqwwq: Nice!

winterfell/webapp/src/js/claimConfirmModifyContactCtrl.js
```diff
@@ -0,0 +1,29 @@
...
+    $scope.targetIndex = modalData.targetIndex;
+    $scope.modalTitle = '';
+    if (modalData.forType == 'address') {
+      $scope.modalTitle = $scope.targetIndex >= 0 ? 'Edit Address' : 'Add Address';
+    } else if (modalData.forType == 'email') {
+      $scope.modalTitle = $scope.targetIndex >= 0 ? 'Edit Email' : 'Add Email';
+    }
+
+    $scope.onCloseModal = function() {
+      $uibModalInstance.close();
```
qwwqwwq: You dont need to make a special function for this, the modal child $scope has certain methods prebound to it by the anguar UI plugin, such that in your modal template you can just say `ng-onclick=close()`

qwwqwwq: Then you can capture a return variable to see if it was accepted, rejected, or dismissed.

winterfell/webapp/src/templates/modals/confirmClaimModifyContact.jade
```diff
@@ -0,0 +1,26 @@
...
+
+  div.input-wrapper(ng-if="forType == 'email'")
+    input(placeholder="Contact Name")
+    input(placeholder="Email")
+    input(placeholder="Confirm Email")
+
+  div.button-wrapper
+    button.cancel-btn(ng-click="onCloseModal()", ng-if="targetIndex < 0") Cancel
+    button.delete-btn(ng-click="onDelete()", ng-if="targetIndex >= 0") Remove
+    button.save-btn(ng-click="onSave()") Save
```
qwwqwwq: Could do something like `ng-click=close({$value: address})` and catch the return value from uibModal.open and then save that return value to the model in the parent controller.

winterfell/webapp/src/templates/modals/confirmClaimModifyContact.jade
```diff
@@ -0,0 +1,26 @@
...
+    input.province(placeholder="State/Province")
+    input.zip(placeholder="Zip/Postal")
+
+  div.input-wrapper(ng-if="forType == 'email'")
+    input(placeholder="Contact Name")
+    input(placeholder="Email")
+    input(placeholder="Confirm Email")
+
+  div.button-wrapper
+    button.cancel-btn(ng-click="onCloseModal()", ng-if="targetIndex < 0") Cancel
```
qwwqwwq: Could do something like `ng-click=close({$value: null})` for this

ahsu1230: done

winterfell/webapp/src/templates/modals/confirmClaimModifyContact.jade
ahsu1230: done

winterfell/webapp/src/js/claimConfirmCtrl.js
ahsu1230: I'm looking for forms that the user has filled out, not all forms. I don't think `formTemplateService` is what I'm looking for.

winterfell/webapp/src/js/claimConfirmCtrl.js
ahsu1230: done. Was going to do that in a future change (linking actual info instead of placeholder items)

ahsu1230:
Still missing a few links such as:
 * Getting all forms a user filled has completed
 * Getting address of a user
 * Editing physical addresses & emails to send copies to

I want this change to just be ui-only. Separate PR will be for linking things correctly (if they are ready).

PR #63, opened 2016-09-17
redo select forms ui only

ahsu1230:
A couple of placeholders, but I think once we have the server & claim/form logic in place, it'll be easy to switch over.
@qwwqwwq


winterfell/webapp/src/js/claimSelectFormsCtrl.js
```diff
@@ -1,13 +1,23 @@
 var app = angular.module('vetafiApp');
-app.controller('claimSelectFormsCtrl', ['$scope', 'claimService', 'formTemplateService', '$stateParams',
-  function($scope, claimService, formTemplateService, $stateParams) {
+app.controller('claimSelectFormsCtrl', ['$scope', 'claimService', '$stateParams',
+  function($scope, claimService, $stateParams) {
     $scope.claimId = $stateParams.claimId;
-    $scope.forms = formTemplateService;
+    $scope.allForms = [
```
qwwqwwq: Why do we need to go back to hardcoded data here? It seems like a step in the wrong direction

winterfell/webapp/src/js/claimSelectFormsCtrl.js
```diff
@@ -1,13 +1,23 @@
...
+      {
+        title: 'VA-2392 Intent to File',
+        summary: 'This form is meant for blah blah something to hold your place in line.',
+        required: true
+      }
+    ];
+    var partitionForms = _.partition($scope.allForms, 'required');
```
qwwqwwq: This logic counting the number of required forms should be encapsulated in a function because we might need to call it multiple times if forms are updated or as a callback

winterfell/webapp/src/js/claimSelectFormsCtrl.js
ahsu1230: These hardcoded values were supposed to just be for placeholders until we solidified the actual dynamic data retrieving.
Do we currently have anything that grabs all forms we support? `formTemplateService` doesn't exist anymore and all I see is a `render/form` endpoint and a `save/form` endpoint.

winterfell/webapp/src/js/claimSelectFormsCtrl.js
ahsu1230: ok, agreed

winterfell/webapp/src/js/claimSelectFormsCtrl.js
ahsu1230: oh I see, `formTemplateService` is in `formly-fields.js`. I got errors because it wasn't being included in the angular app correctly. Maybe because it is on the same level as `index.jade` / `index.html`? Was that on purpose?

qwwqwwq: It is at the top level because it is just a placeholder that imports the forms via browserify. Because it is the only thing that uses browserify i wanted to separate it out

qwwqwwq: It should be bundled with the rest of the code when you run `gulp build`, has been working fine for me, is that not true for you?

ahsu1230: Yea, it works only if I use vagrant. Just using gulp to do front-end only development won't work - the app can't find formTemplateService so angular crashes. :/
So this kind of makes front-end dev harder

winterfell/forms/VBA-21-0966-ARE.json
```diff
@@ -1,4 +1,8 @@
 {
+  "unofficialTitle": "Intent to File",
```
ahsu1230: @qwwqwwq am I allowed to do this? I just hate the official name and description pulled from the form. I think if we want things to be more presentable, we'll have to redefine some fields.
But, since this file is auto-generated, is there a better place to add fields?
We may have to look into a pipeline where we combine autogenerated stuff with additional fields we define to create this file.

winterfell/webapp/src/templates/form.jade
```diff
@@ -1,4 +1,4 @@
-div.form-wrapper
+div#vfi-form-view
```
ahsu1230: I was using a class `form-wrapper` and was getting unexpected properties. To avoid this, I could either find another class name or make this class an id instead. To be consistent with the rest of the app, I think this would be best with an id.

winterfell/webapp/src/styles/form.styl
```diff
@@ -30,19 +30,19 @@
...
     right: 0;
 
-.form-wrapper
+#vfi-form-view
   margin-left: 10%
   margin-right: 10%
 
-.form-submit-button-group
-  margin-top: 3vh
+  & .form-submit-button-group
```
ahsu1230: when you move all these classes under the parent id,
you can remove confusion and mixups with all `.form-title`, `.form-submit-button-group`, `.form-name`, etc.
You can know for sure that all of those classes are used in this page and only this page.

winterfell/webapp/src/js/claimSelectFormsCtrl.js
```diff
@@ -1,13 +1,65 @@
 var app = angular.module('vetafiApp');
-app.controller('claimSelectFormsCtrl', ['$scope', 'claimService', 'formTemplateService', '$stateParams',
-  function($scope, claimService, formTemplateService, $stateParams) {
+app.controller('claimSelectFormsCtrl', ['$scope', '$state', '$stateParams', 'claimService', 'formTemplateService',
+  function($scope, $state, $stateParams, claimService, formTemplateService) {
     $scope.claimId = $stateParams.claimId;
-    $scope.forms = formTemplateService;
+
+    $scope.allForms = formTemplateService; // returns an object {formId -> {...}}
```
ahsu1230: just a note. For some reason, `formTemplateService` is not found by angular when you just run
`gulp` (not `gulp build`).
Running gulp creates a local webserver to host your webapp and will auto-update whenever you change a css property, template, js, etc so you can view your changes without having to reload pages.
`gulp build` just does all the proprocessing stuff and moves those files to the `build` folder.

qwwqwwq: Hmm that is odd, that javascript is generated by the `browserify` step which is a requirement of the `build` step which is part of the `default` step, so just running `gulp` *should* be doing the right thing.

winterfell/forms/VBA-21-0966-ARE.json
qwwqwwq: I agree, those official names were just a placeholder, I assumed we would go back and make them more user friendly eventually. This file is NOT autogenerated, so you are free to add fields, as long as that same field will exist for all the forms for consistency.

winterfell/forms/VBA-21-0966-ARE.json
```diff
@@ -1,4 +1,8 @@
...
+  "unofficialDescription": "This form is used to notify VA of your intent to file for benefits. Even if your claim fails to process, your benefits effective date will still be the date the VA receives this form.",
+  "required": true,
```
qwwqwwq: required for what? we should define this better eventually.

winterfell/webapp/src/styles/claimSelectForms.styl
```diff
@@ -1,97 +1,187 @@
...
+  margin: 0 auto
+  padding: largeSpacing 0
+
+  & h3
+    margin-bottom: mediumSpacing
+    color: colorPrimaryBlue
+
+  & h4
+    display: inline-block
+    line-height: 40px
```
qwwqwwq: nervous about these hardcoded text sizes, cant we just use `em` or something more mobile friendly?

winterfell/webapp/src/styles/claimSelectForms.styl
```diff
@@ -1,97 +1,187 @@
...
+      vertical-align: top
+      text-align: right
+
+      @media screen and (max-width: maxWindowWidthPx)
+        width: 100%
+        text-align: center
+
+      & button
+        display: inline-block
+        width: 150px
```
qwwqwwq: hardcoded pixel size

winterfell/webapp/src/styles/claimSelectForms.styl
```diff
@@ -1,97 +1,187 @@
...
+          background: url('../icons/warning-white.svg')
+          background-repeat: no-repeat
+          background-size: cover
+
+      &.complete
+        background-color: colorLightGreen
+
+        & .state-icon
+          transform: translate(-50%, -50%)
+          width: 22px
```
qwwqwwq: im more ok with the hardcoded icon size, but could be a problem on very small screens, maybe we can think about using media queries or just make them flat out responsive.

winterfell/webapp/src/templates/claimSelectForms.jade
```diff
@@ -1,15 +1,34 @@
...
-    div.vfi-button-wrapper
-      button.vfi-cancel-btn(ng-click="onClickCancel()") Cancel
-      button.vfi-finish-btn(ui-sref="root.claimconfirm({claimId: claimId})") Finish
+  #select-forms-wrapper
+    h3 Select Forms for Claim
+
+    div.section.required
+      div.section-header
+        h4 Required Forms:
+        uib-progressbar.progressbar(value="numRequiredCompleted", max="requiredForms.length", type="{{progressBarType}}")
```
qwwqwwq: Just a note im eventually going to rig this up to show the percent completion in terms of questions answered, and not just in terms of forms completed.

winterfell/webapp/src/js/claimSelectFormsCtrl.js
```diff
@@ -1,13 +1,65 @@
...
+    var userForms = [ // TODO: state of all forms edited by user
```
qwwqwwq: I'm not sure what you need this hardcoded object for, it seems like you just use it to iterate the possible forms that could be completed. You could just iterate through the keys in `formTemplateService` to the same effect, and that will avoid the hardcoding

ahsu1230: On the general idea, I would like to know out of forms [A,B,C,D,E], which forms have been edited by a user.

Maybe the user finished A and D and left C unfinished. Forms B and E are untouched.
So the list of ALL forms and the list of forms edited by a user is different.

In this case, `userForms` could be
```
[ { id: A, state: 'complete' }, { id: C, state: 'unfinished'}, { id: D, state: 'complete' } ]
```

`formTemplateService` gives us all forms that can be completed. But it doesn't give us which forms are completed, unfinished or untouched. The unfinished state is important because if I think it would be helpful to know if a form is partially completed, but was meant to be completed. From here, a user can discard it or edit it to complete the form. For now, since we only have one form, we can go ahead and just have two states: `completed` and `incomplete`.

But it seems to me `formTemplateService` is mainly for immutable values. We need a concept of mutable `state` values for each form.

qwwqwwq: That is correct, the `formTemplateService` is supposed to be immutable, given a form id, it gives you the template which allows you to construct the list of questions and present them to the user (the name `formTemplateService` might be bad, but its just what I randomly thought to call it).

What you want is the state of current forms. I feel like that is something that should be stored in the database and retrieved for a given claim id/user id combo.

ahsu1230: Ok great, naming is fine. And I think we're on the same page then. 
`formTemplateService` should continue to be used as an immutable list for all forms.

As for the state of current forms, we don't have it at the moment so for now, please let me keep this hardcoded variable so at least there's some UI present. When we're ready to hook things up in a future change, this hardcoded `userForms` will be gone. These last few changes  from me have been focusing on what we need for our MVP UI-wise anyway. There's a little more backend to be done anyway, but i want the hookup with the front-end to be simple.

winterfell/forms/VBA-21-0966-ARE.json
ahsu1230: This form is required for submitting a valid claim. If you do not complete this/these required form(s), you cannot submit a claim.

winterfell/webapp/src/styles/claimSelectForms.styl
ahsu1230: ok will change

winterfell/webapp/src/styles/claimSelectForms.styl
ahsu1230: you're right here. Line-heights for headers are defined in `em` anyway. I was being inconsistent here.

winterfell/webapp/src/styles/claimSelectForms.styl
ahsu1230: So far things have been scaling well. I test out responsiveness by adjusting window width and/or increasing the zoom to 200% (I think that's an actual guideline in the U.S. Web Design Standards). Even using Chrome's Inspect Device tool has been working okay. In which case, most mobile users would zoom in anyway and recent browsers have been getting alot better at scaling in px. So I think icon sizes in px are fine. If we ever get to actually viewing it on phones, we can revisit this then.

Icon sizes are more particular in how big they need to be, and I don't want to be specifying icon sizes abstractly like this: `width: 1.34em` throughout the entire codebase.

ahsu1230:
Ok found a middle ground for us. The `userForms` will be an empty object (changed to object so it works similarly to `formTemplateService`) and in case I want a quick reference of what it looks like UI wise, I can uncomment the lines right next it so I can see how it looks when the form is complete.

PR #64, opened 2016-09-17
breadcrumbs directive used in multiple pages

ahsu1230:


winterfell/webapp/src/templates/breadcrumbs.jade
```diff
@@ -0,0 +1,5 @@
+#vfi-breadcrumbs
+  div.breadcrumbs-wrapper
+    div.link-wrapper(ng-repeat="link in links")
+      span(ng-class="{'lighten': link.lighten}", ng-if="$index > 0") >
```
qwwqwwq: Looks like theres a typo here

ahsu1230: I want it to display a `>`. Maybe i should be using an html encoded character

qwwqwwq: Oh i see, yeah not sure

ahsu1230:
fixed typo

PR #65, opened 2016-09-17
remove claim start page

ahsu1230:
Our MVP is only for filing an intent form.
I believe that a start-claim page is kind of useless.
We should just skip and go straight to selecting forms to fill out.


winterfell/webapp/src/js/claimStartCtrl.js
```diff
@@ -1,22 +0,0 @@
-var app = angular.module('vetafiApp');
-app.controller('claimStartCtrl', ['$scope', '$state', 'net', 'claimService', '$uibModal',
-  function ($scope, $state, net, claimService, $uibModal) {
-    function getTosAgreement() {
```
qwwqwwq: When are we going to ask them to fill out the TOS?

ahsu1230:
After thinking about it. I think there is still value with a start page.
It would be a good place to give an intro as to why filling out the Intent to File form is important.
And the ToS agreement too. Didn't think that through lol.

PR #70, opened 2016-09-18
Add Validator for "State" Field

qwwqwwq:
This validator will make sure the state is two capital letters, will not let the user type any other characters in the form field.


ahsu1230:
lgtm

PR #78, opened 2016-09-25
signup-login tweaks

ahsu1230:
* fixed height
* got back button to show up again (when click on Login/Signup with email)
* use 1 file `signin.jade` instead of having two files. They were too similar anyway
* pressing enter should trigger the form submit


winterfell/webapp/src/js/noangular/signin.js
```diff
@@ -36,7 +38,30 @@ function displayError(msg) {
   $('.error-msg').text(msg);
 }
 
+function adjustTextsForLogin() {
```
qwwqwwq: This feels a little weird to me having one page and then editing the DOM to make it look like either signup or login. I'm afraid it might make it hard to change either of these pages. I think it might be a lot nicer to use jade/pug's inheritance features:

https://pugjs.org/language/inheritance.html

We could have one base signin page, and then have two different blocks with either the fields for creating an account or the fields for logging in

winterfell/webapp/src/js/noangular/signin.js
```diff
@@ -49,15 +74,15 @@ $(document).ready(function(){
     $('.inputs').removeClass('vfi-show');
   });
 
-  $('#signup-view .submit-btn').click(function(e) {
+  $('#signup-view .inputs form').submit(function(e) {
```
qwwqwwq: Nice, my understanding is we want login/signup to be as close to a vanilla html form as possible. Could we maybe just use the `<form onsubmit=` attribute instead of jquery? I think that will help some browsers. This page is pretty simple we might not need jquery at all.

winterfell/webapp/src/js/noangular/signin.js
ahsu1230: sweet, used the jade inheritance. it works well!

winterfell/webapp/src/js/noangular/signin.js
ahsu1230: I'm actually going to keep it with jquery.
There are a lot of things that are just easier with jquery like a POST request, grabbing values from elements, setting text for certain elements, adding/removing classes, etc.
Since most sites use jQuery anyway, I don't think it's super critical to optimize to the point of using vanilla over jQuery.

PR #79, opened 2016-09-25
Redo start claim page

ahsu1230:
@qwwqwwq


winterfell/webapp/src/js/tosCtrl.js
```diff
@@ -13,6 +13,7 @@ app.controller('tosCtrl', ['$scope', '$location', 'claimService',
     };
 
     $scope.onDecline = function() {
+      $scope.$close(true);
```
qwwqwwq: What does the `true` argument mean?

winterfell/webapp/src/styles/claimStart.styl
```diff
@@ -1,22 +1,89 @@
...
 
+maxContentWidthPx = 560px
+iconWrapperPx = 54px
+
 #claim-start-view
-  width: maxLayoutWidthPx
-  margin: largeSpacing auto
-  padding: largeSpacing megaSpacing
-  box-sizing: border-box
+  min-height: 600px
```
qwwqwwq: why do we need min height? could we make it in units percent or `vh` instead of pixels if we do need it?

qwwqwwq:
Overall LGTM just a few questions

winterfell/webapp/src/styles/claimStart.styl
ahsu1230: this was a cosmetic thing because the content itself here isn't a lot so you see the footer pretty high up.
I just chose a height big enough to make it so it's not too low (too much empty space) and not too high.

winterfell/webapp/src/js/tosCtrl.js
ahsu1230: oh I just copied it from line 8. I thought it just closed the modal.

PR #80, opened 2016-09-28
Two generic modals for simple messaging

ahsu1230:
Two generic modals:

A) One modal with one button. This is like a FYI modal. There will be a message and when a user exits out of this modal, one action is done after the user closes the modal.

I replaced the Session Expired modal with this modal since there is only one action you go to after pressing OK, back to the login screen.

B) Another modal with two buttons. This is like a Yes/No modal. There are two buttons: one for continuing, and another for canceling. Depending on which button the user selects, the appropriate action is done. You can also style the continue button using a `choice` variable. This helps give you more control of what kind of messaging you want to give.

I replaced the Delete Account modal with this modal, since a user can choose to confirm that they want to delete their account or cancel the action.

I wanted to make this as easy to use as possible, so I avoided creating two new controllers just for these generic modals.

### DIFF
--- a/winterfell/webapp/src/js/headerCtrl.js
+++ b/winterfell/webapp/src/js/headerCtrl.js
@@ -32,19 +32,19 @@ app.controller("headerCtrl",
 
         }, function failure(res) {
           that.active = true;
-          var modal = $uibModal.open({templateUrl: 'templates/modals/sessionExpired.html', windowClass: 'ngdialog-theme-default'});
-          modal.result.then(
-            function success() {
-              console.log("success");
-              that.active = false;
-              $window.location.href = '/login';
-            },
-            function failure() {
-              console.log("fail");
-              that.active = false;
-              $window.location.href = '/login';
-            }
-          );
+          $scope.modalData = {
+            headline: "Session Expired",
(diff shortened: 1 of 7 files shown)


winterfell/webapp/src/js/headerCtrl.js
```diff
@@ -32,19 +32,17 @@ app.controller("headerCtrl",
...
-            function failure() {
-              console.log("fail");
-              that.active = false;
-              $window.location.href = '/login';
-            }
-          );
+          $scope.headline = "Session Expired";
+          $scope.message = "For security reasons, your session has expired due to inactivity. Please log back in to continue your work.";
+          var modal = $uibModal.open({
+            scope: $scope,
```
qwwqwwq: Instead of polluting the parent scope, might be cleaner to pass `scope: {headline: ..., message: ...}`

ahsu1230: done. I tried before doing the inline scope object
```
scope: { headline: ..., message: ... }
```

but it gave me errors and seems like you have to actually create new scope objects using `$scope.$new()`. Now it works as intended.

PR #82, opened 2016-09-29
Make Biscuit Synchronous

qwwqwwq:
Make biscuit synchronous so that the global app object contains them before the app starts. This fixes the problems with the unit tests.


winterfell/app.js
```diff
@@ -32,6 +42,7 @@ loadIntoBuild(app, 'middlewares');
 loadIntoBuild(app, 'services');
 loadIntoBuild(app, 'controllers');
 
+
```
ahsu1230: nit: extra line

ahsu1230:
lgtm

PR #83, opened 2016-09-30
Store Form Progress In DB

qwwqwwq:
- Stores number of answered/answerable questions along with each form so we can show total claim progress
- Calculate progress of each form on the backend by executing business logic of which questions are answerable/answered. We use the same logic we use on the frontend to tell the user in realtime their progress by using the angular-expressions library in the backend to evaluate the same conditionals that determine whether a question needs to be answered.
- The claim creation endpoint is now a post that takes a list of forms, and those forms will be initialized along with the claim
- The claim select page now resolves the forms via the ui router so that the progress can be displayed (haven't updated the template to actually display it yet though).

### DIFF
--- a/winterfell/models/form.js
+++ b/winterfell/models/form.js
@@ -8,7 +8,9 @@ var formSchema = new Schema({
   key: String, // Code name for the form
   responses: Schema.Types.Mixed,
   user: Schema.Types.ObjectId,
-  claim: Schema.Types.ObjectId
+  claim: Schema.Types.ObjectId,
+  answered: Schema.Types.Integer,
+  answerable: Schema.Types.Integer
 }, {
   timestamps: true
 });
--- a/winterfell/models/form.js
+++ b/winterfell/models/form.js
@@ -9,8 +9,8 @@ var formSchema = new Schema({
   responses: Schema.Types.Mixed,
   user: Schema.Types.ObjectId,
   claim: Schema.Types.ObjectId,
-  answered: Schema.Types.Integer,
-  answerable: Schema.Types.Integer
+  answered: Number,
+  answerable: Number
 }, {
   timestamps: true
 });
--- a/winterfell/services/claimService.js
+++ b/winterfell/services/claimService.js
@@ -12,6 +12,8 @@ function ClaimService(app) {
   this.app = app;
 }
 
+module.exports = ClaimService;
+
 /**
  * For a given form and set of responses, calculate how many questions
  * were answered (which is simply the number of keys in the responses),
@@ -19,13 +21,16 @@ function ClaimService(app) {
  * which requires evaluating the angular hideExpression attribute for
  * each field against the current responses.
  *
- * @param formName
+ * @param formId
  * @param data
  * @returns {{answerable: number, answered: number}}
  */
-function calculateProgress(formName, data) {
+module.exports.calculateProgress = function calculateProgress(formId, data) {
   var evaluate, i;
-  var template = formlyFields[formName];
+  if (!(formId in formlyFields)) {
+    throw new Error("Unknown formId: " + formId);
+  }
+  var template = formlyFields[formId];
   var output = {answerable: 0, answered: _.size(data)};
 
   if (!template) {
@@ -46,11 +51,8 @@ function calculateProgress(formName, data) {
   }
 
   return output;
-}
+};
 
-module.exports.calculateProgress = calculateProgress;
-
-module.exports = ClaimService;
 module.exports.findIncompleteClaimOrCreate = function(userId, forms, callback) {
   return Claim.findOne({ userId: userId, state: Claim.State.INCOMPLETE }).exec(function(err, fileClaim) {
     if (err) {
--- a/winterfell/services/claimService.js
+++ b/winterfell/services/claimService.js
@@ -4,12 +4,51 @@ var httpErrors = require('./../utils/httpErrors');
 var Claim = require('./../models/claim');
 var Form = require('./../models/form');
 var Q = require('q');
-var mongoose = require('mongoose');
-mongoose.Promise = Q.Promise;
+var bulk = require('bulk-require');
+var formlyFields = bulk(__dirname + '/../forms/', ['*']);
+var expressions = require("angular-expressions");
 
 function ClaimService(app) {
   this.app = app;
 }
+
+/**
+ * For a given form and set of responses, calculate how many questions
+ * were answered (which is simply then number of keys in the responses),
+ * and more complicatedly, how many questions on the form were answerable,
+ * which requires evaluating the angular hideExpression attribute for
+ * each field against the current responses.
+ *
+ * @param formName
+ * @param data
+ * @returns {{answerable: number, answered: number}}
+ */
+function calculateProgress(formName, data) {
+  var evaluate, i;
+  var template = formlyFields[formName];
+  var output = {answerable: 0, answered: _.size(data)};
+
+  if (!template) {
+    output.answerable = null;
+    return output;
+  }
+
+  for (i = 0; i < template.fields.length; i++) {
+    var field = template.fields[i];
+    if (field.hasOwnProperty('hideExpression')) {
+      evaluate = expressions.compile(field.hideExpression);
+      if (!evaluate({model: data})) {
+        output.answerable += 1;
+      }
+    } else {
+      output.answerable += 1;
+    }
+  }
+
+  return output;
+}
+
+module.exports.calculateProgress = calculateProgress;
 
 module.exports = ClaimService;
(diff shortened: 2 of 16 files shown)


winterfell/services/claimService.js
```diff
@@ -2,18 +2,88 @@ var _ = require('lodash');
...
+ * which requires evaluating the angular hideExpression attribute for
+ * each field against the current responses.
+ *
+ * @param formName
+ * @param data
+ * @returns {{answerable: number, answered: number}}
+ */
+function calculateProgress(formName, data) {
+  var evaluate, i;
+  var template = formlyFields[formName];
```
ahsu1230: we want to be careful with what `formName` vs. `formId` means. I think there might be some inconsistencies in a few places (I've done it as well).

To me,
`formId` = `VBA-21-0966-ARE1`
`formRowId` = mongoDb row id of form
`formName` = "INTENT TO FILE A CLAIM FOR COMPENSATION AND/OR PENSION, OR SURVIVORS PENSION AND/OR DIC" (as seen in `VBA-21-0966-ARE.json`) We could also call this formTitle

`formUnofficialName` = "Intent to File" (our easier-to-read form title). We could also call this `formUnofficialTitle` or maybe something like `form.vfi.name` to convey that everything in the object `vfi` is our stuff.
Thoughts?

winterfell/test/testClaimController.js
```diff
@@ -210,9 +211,40 @@ describe('SaveClaimController', function () {
...
         Form.findOne({key: '1', user: targetUser._id}, function(error, doc) {
           should.not.exist(error);
+          should.exist(doc);
           doc.responses.should.deepEqual({key1: 'value1', key2: 'value2'});
           done();
         })
       });
   });
+
+  it('Should correctly calculate progress after save', function(done) {
```
ahsu1230: It's ok to leave this test here for now.
Some somewhere down the line, if we want to test individual forms (like VBA-21-0966 here), we should have one separate test file for testing individual forms. `testClaimController` should just test your general basic claim functions

winterfell/services/claimService.js
```diff
@@ -2,18 +2,88 @@ var _ = require('lodash');
...
+  var output = {answerable: 0, answered: _.size(data)};
```
ahsu1230: _.size(data) as answered could be a dangerous. Let's say we have 5 questions A, B, C, D, E and only A and B are important and required.
If I just answer C and D, `answered` would count those two even though we haven't really completed the form.

Maybe you've gone about things in which every question that is `answerable`, is required. Is that the case? I think there could be a case where you could choose to answer a question, but if you skipped it, it's okay so it's technically optional.

ahsu1230:
nice work so far. A lot of it is just inconsistency questions and what `answerable` means.

winterfell/models/form.js
```diff
@@ -8,7 +8,9 @@ var formSchema = new Schema({
   key: String, // Code name for the form
   responses: Schema.Types.Mixed,
   user: Schema.Types.ObjectId,
-  claim: Schema.Types.ObjectId
+  claim: Schema.Types.ObjectId,
+  answered: Number,
+  answerable: Number
```
qwwqwwq: I think it makes forms a bit more complicated than necessary to divide things into "required" and "optional",  my thought for vetafi is there there will never be any optional questions, which makes it easier for the user, because if you see it, you need to answer it. 

The methods for addressing pieces of information that may or may not be necessary then becomes asking gating questions, and having the "optional" questions become visible depending on the answer to those gating questions. For instance we can ask "Do you have a VA file number?", and then if you answer yes, the question "What is your VA file number?" becomes visible. 

For things like address line 2, we can just have the line being blank be considered "complete" in validation.

What do you think?

winterfell/webapp/src/js/net.js
```diff
@@ -59,14 +59,17 @@ app.factory('net', ['xhrEnv', '$http', function(xhrEnv, $http) {
...
+    startClaim: function(data) {
+      return httpPost("/claims/create", data);
     },
     submitClaim: function(extClaimId) {
       return httpPost("/claim/" + extClaimId + "/submit");
     },
     discardClaim: function(extClaimId) {
       return httpDelete("/claim/" + extClaimId);
+    },
+    getFormsForUser: function(extClaimId) {
```
qwwqwwq: agreed

winterfell/webapp/src/js/claimStartCtrl.js
```diff
@@ -14,7 +14,9 @@ app.controller('claimStartCtrl', ['$scope', '$state', 'net', 'claimService', '$u
           return;
         }
       }
-      net.startClaim().then(function (res) {
+      net.startClaim(
+        {forms: ["VBA-21-0966-ARE"]} // hardcoded to create claim with only this form for now
```
qwwqwwq: That sounds reasonable, basically I was thinking that somehow the set of forms should be determined by the type of claim the person chooses to file, and we would give them certain "pre bundled" options or something. I'd like to leave it for a future revision for now though I'm not sure what to do here atm

winterfell/test/testClaimService.js
```diff
@@ -5,26 +5,30 @@ var Claim = require('../models/claim');
...
-         state: User.State.ACTIVE
-       };
-       User.create(user, function(err, user) {
-         targetUser = user;
-         done();
-       });
-     });
-   });
+  before(function (done) {
+    this.timeout(10000);
```
qwwqwwq: Oh I dont know why I set this, the test certainly doesnt take 10 seconds, will remove

winterfell/services/claimService.js
qwwqwwq: That sounds good to me, slightly prefer formTitle over formName since its kind of long for a name..

winterfell/test/testClaimController.js
qwwqwwq: Oh I was not even thinking on that level, I was just having `testClaimController` is just testing all the endpoints in the `claimController.js` file. I think maybe we need to move the form logic into a different controller package if this one is becoming too cluttered.

winterfell/services/claimService.js
qwwqwwq: This logic is really just a catch all if the formId that is supplied does not exist or is invalid. I figured it was the most reasonable behavior considering, but now I think its kind of Janky.

I think this method should just throw if the formId is unknown.

winterfell/models/form.js
ahsu1230: ok I see. I think that works pretty well for now. I think it's okay to have some things already considered "complete" even if it's blank.

winterfell/webapp/src/js/app.js
```diff
@@ -96,7 +96,22 @@ app.config(['$stateProvider', '$urlRouterProvider', function ($stateProvider, $u
     name: 'root.claimselect',
     url: '/claim/{claimId}/select-forms',
     templateUrl: 'templates/claimSelectForms.html',
-    controller: 'claimSelectFormsCtrl'
+    controller: 'claimSelectFormsCtrl',
+    resolve: {
+      forms: ['net', '$q', '$stateParams', function(net, $q, $stateParams) {
+        var deferred = $q.defer();
+
+        net.getFormsForClaim($stateParams.claimId).then(
```
ahsu1230: just a question. Why do this here in this resolve? As opposed to calling it when the controller starts?

winterfell/webapp/src/js/claimSelectFormsCtrl.js
```diff
@@ -1,55 +1,17 @@
...
-        targetForm.id = formId;
-        targetForm.state = calcFormState(userFormState, targetForm.required);
-
-        if (targetForm.required && targetForm.state == 'complete') {
-          numRequiredCompleted++;
-        }
-      });
-
-      $scope.numRequiredCompleted = numRequiredCompleted;
-      $scope.progressBarType = $scope.numRequiredCompleted == $scope.requiredForms.length ? 'success' : undefined;
```
ahsu1230: `$scope.progressBarType` is missing. Logic is pretty simple. Could you add this line maybe after the line below?

ahsu1230:
A question and one comment. Otherwise LGTM!

winterfell/webapp/src/js/app.js
qwwqwwq: I like the resolve feature for two reasons:

- Child states inherit the resolved objects, which lets us avoid rewriting all the code to make API requests.
- Having the result of the resolve injected into the controller will make it easier to unit test the controllers when we eventually write frontend unit tests.

qwwqwwq:
Ok updated with the progress bar type, but I put the logic into the view instead of in the controller.

PR #84, opened 2016-10-03
Redo profile page

ahsu1230:


winterfell/webapp/src/js/profileCtrl.js
```diff
@@ -1,54 +1,48 @@
...
-      },
-      {
-        rank: 'captain',
-        branch: 'army',
-        yearStart: '2012',
-        yearEnd: '2013'
-      }
-    ];
+    $scope.user = Profile.user.user;
+    $scope.claims = []; // list of user's claims
```
qwwqwwq: You can resolve claims through the UI router and then inject them into the controller, similar to how we do for other things like the user and forms, etc.

winterfell/webapp/src/templates/profile/claims.jade
```diff
@@ -1,18 +1,33 @@
...
-                    div.vfi-icon.vfi-view
-        div.vfi-check-icon(ng-if="claim.state=='submitted' || claim.state=='processed'")
-        div.vfi-bar
-            div.vfi-progress(ng-class="{'vfi-incomplete': claim.state=='incomplete', 'vfi-submitted': claim.state=='submitted', 'vfi-processed': claim.state=='processed'}")
+div.profile-section#claims
+
+  div.no-claims-wrapper(ng-show="claims.length == 0")
+    h4 You have no pending file claims at the moment.
+    br
+    p Click #[a(ui-sref="root.claimstart") here] to start a file claim.
```
qwwqwwq: so in jade/pug the `[` denoes an inline HTML element?

ahsu1230: yup! http://stackoverflow.com/questions/6989653/jade-links-inside-a-paragraph
shows you how to do inline elements

winterfell/webapp/src/js/profileCtrl.js
ahsu1230: What's the difference between doing that and using `claimService.userClaims` instead?

PR #85, opened 2016-10-05
redo view claim page (ui)

ahsu1230:
Now we can grab claim data and which forms were associated with that claim.

TODO:
every claim should have a `recipient` object or two fields: `emails` and `addresses`
In this view, it would be helpful to know where these claims were sent to.


ahsu1230:
a bit of plumbing needs to happen first

winterfell/webapp/src/js/profileCtrl.js
```diff
@@ -36,7 +15,9 @@ app.controller('profileCtrl', ['$scope', '$location', '$window', 'Profile', 'cla
 
     function init() {
       for (var i = 0; i < $scope.claims.length; i++) {
+        $scope.claims[i].id = $scope.claims[i].externalId; // todo, server shouldn't be sending rowId and externalId. We only reference claims through claim.id
         $scope.claims[i].header = createHeaderString($scope.claims[i]);
+        $scope.claims[i].date = utils.formatServerDate($scope.claims[i].stateUpdatedAt);
```
qwwqwwq: You could use the pattern in `formCtrl.js`, which is to inject `$filter` and then call `$filter('date')(new Date(), 'MM/dd/yyyy');`, this uses angulars built in date formatter

ahsu1230: oh nice

PR #86, opened 2016-10-08
Refactor Document Rendering / Mailing

qwwqwwq:
- Removed unnecessary models (DestinationAddress, Document) in favor of just adding a pdf field to Form which is really all we need.
- Get rid of the weird POST/REDIRECT/GET thing that we were using to render forms in response to request, instead just update the pdf each time the form is saved, and just the PDF from that mongo document directly when needed.
- Get address from UserValues for claimSubmit page
- Update tests


qwwqwwq:
I noticed this has some conflicts with the claimSelect page, I tried to simplify it to just be based off the direct set of Form model objects returned from mongoose. Will need to take a closer look at these conflict now to incorporate the features you were trying to add to the page.

winterfell/webapp/src/js/claimConfirmCtrl.js
```diff
@@ -1,87 +1,87 @@
...
-    ];
+    $scope.formsList = forms;
+    $scope.formsInfo = formTemplateService;
 
-    $scope.onClickAddAddress = function() {
-      return openModifyContactModal('address', $scope.mailingList);
-    };
+    console.log($scope);
 
-    $scope.onClickEditAddress = function(index) {
```
ahsu1230: thanks for fixing this part. We should only edit address/email (and it should reflect on the form?) and not be adding more addresses/emails.

winterfell/webapp/src/js/claimConfirmCtrl.js
```diff
@@ -1,87 +1,87 @@
...
     }
 
-    $scope.onClickConfirm = function() {
-      var claim = claimService.getIncompleteClaim();
-      if (!_.isEmpty(claim)) {
-        net.submitClaim($stateParams.claimId).then(function(resp) {
+    $scope.onClickConfirm = function () {
+      net.submitClaim($stateParams.claimId,
+        {
+          toAddress: $scope.vaAddress,
```
ahsu1230: technically we are sending to two places: the VA, and a copy to the user's address. Then we are emailing a copy to the user's email.

Now that I think about it, as a service, when we send a letter to the VA, I'm not sure the return address should be a user's address. It should be us. If we screwed up somehow, the letter should go back to us so we can fix it. Having it go back to the user doesn't help since he/she is less likely able to do stuff about it AND already has a copy anyway. If we need the user to fix something, we should email them through our app with instructions instead of hoping that they will fix things themselves.

The return address for both of the letters should be Vetafi since we are acting as a middle ground between the user and the VA. When we send a letter to the VA, the title could even be something like: "From: Vetafi: Aaron Hsu" if we really wanted to show on the letter who it is for. Same thing for a user. If we send a copy to the user, they should know it's from us.

So from this perspective, this body could just be:
```
net.submitClaim(claimId,
  {
    addresses: [vaAddress, userAddress],
    emails: [user.email]  
  }
)
```
This makes the endpoint generic (as mentioned in an earlier comment) since we just have a list of addresses and emails and whatever letters/emails we send, all parties know it's from us.

winterfell/controllers/claimController.js
```diff
@@ -7,10 +7,16 @@ var User = require('../models/user');
 var Form = require('../models/form');
 var UserValues = require('../models/userValues');
 var auth = require('../middlewares/auth');
+var MailingService = require('./../services/mailingService');
+var DocumentRenderingService = require('./../services/documentRenderingService');
 var _ = require('lodash');
+var Q = require('q');
```
qwwqwwq: Agreed. There are linters out there which do this automatically, we should pick one and make it into a precommit hook eventually.

winterfell/controllers/claimController.js
```diff
@@ -94,15 +100,55 @@ module.exports = function (app) {
...
+  app.post('/claim/:extClaimId/submit',  auth.authenticatedOr404, function(req, res) {
     console.log('[submitClaim] request received ' + JSON.stringify(req.body));
+    var that = this;
+    var currentUser = null;
+    var promise = User.findById(req.session.userId);
 
-    handleClaimStateChange(req.params.extClaimId,
-      Claim.State.SUBMITTED,
-      claimUpdateCallbackFactory(res));
+    promise = promise.then(function(user) {
```
qwwqwwq: yes exactly

winterfell/controllers/claimController.js
```diff
@@ -94,15 +100,55 @@ module.exports = function (app) {
...
+      currentUser = user;
```
qwwqwwq: Yeah I need to go back and standardize handling of that across all the controllers, we haven't been too strict about it so far.

winterfell/webapp/src/js/claimConfirmCtrl.js
qwwqwwq: This is an important discussion, and complex issue, we should move this to a google doc to record thoughts. Receiving mail on behalf of people seems thorny from a legal perspective..

winterfell/services/claimService.js
```diff
@@ -21,20 +21,16 @@ module.exports = ClaimService;
...
  * @returns {{answerable: number, answered: number}}
  */
-module.exports.calculateProgress = function calculateProgress(formId, data) {
+function calculateProgress(formName, data) {
   var evaluate, i;
-  if (!(formId in formlyFields)) {
-    throw new Error("Unknown formId: " + formId);
-  }
-  var template = formlyFields[formId];
+  var template = formlyFields[formName];
```
ahsu1230: hmm I think we said:
```
formId: “VBA-21-0966-ARE”,
formTitle: "INTENT TO FILE A CLAIM FOR COMPENSATION AND/OR PENSION, OR SURVIVORS PENSION AND/OR DIC",
vfi: {
   title: "Intent to File"
   description: "..."
}
```

but then it's possible a form that is saved in db has an _id (row id). We could change `formId` to `formName`, but we'll need to differentiate that from `formTitle`

winterfell/webapp/src/js/claimConfirmCtrl.js
qwwqwwq: I think having the service `sendLetter` function mail to one single address is a simple and most reusable interface. In our application logic we can easily call the `sendLetter` function twice to mail to two separate addresses. I agree I haven't implemented the "send copy" functionality, but I'd like to address that in a future PR because this one is getting way too large and its starting to block my progress because I have to rebase my stuff off of here all the time since master has gotten far behind.

ahsu1230: ok sounds good

ahsu1230:
you have some merge conflicts. One comment I have is just the formId vs. formName.
Other than that, I think we should just merge this.

PR #87, opened 2016-10-09
Change password ui + server hookup

ahsu1230:
- Change Password js logic is through a directive since we are manipulating and extracting information from the dom
- `ChangePasswordModal` inherits a generic modal layout. `OneButtonModal` and `TwoButtonModal` also inherit from this general modal layout. 
- Server-side change password is very simple. It takes in an old password that we check to make sure it matches the current password and sets the new password if things are correct.
- unit test for changing password


ahsu1230:
A big security issue right now is we store passwords in mongoDb as is. And we send password strings in POST bodies as is. In the future, we should store only hashed passwords and any network communication from the client should go through some sort of client-side hashing. But, in the future!

ahsu1230:
I know this feature isn't super prioritized, but it was a simple example to demonstrate (mainly for myself) how to refactor all modals to inherit from a generic layout and able to be manipulated through a directive.

qwwqwwq:
Definitely agree that we should store passwords in the database hashed. However I disagree with the point about hashing passwords client side, the web-standard is to use HTTPS, which will encrypt all the content in the POST request.

See https://www.owasp.org/index.php/Broken_Authentication_and_Session_Management under `Protecting Credentials in Transit`:

> The only effective technique is to encrypt the entire login transaction using something like SSL. Simple transformations of the password such as hashing it on the client prior to transmission provide little protection as the hashed version can simply be intercepted and retransmitted even though the actual plaintext password might not be known.

winterfell/webapp/src/templates/modals/changePassword.jade
```diff
@@ -1,6 +1,16 @@
...
-  input(placeholder='Old password')
-  input(placeholder='New password')
-  input(placeholder='Confirm New password')
+extends modalLayout.jade
+
+block modalContent
+  div#vfi-modal-change-password(modal-change-password)
+    form
+      h4 Old Password
+      input.old(placeholder='Enter your current password', type='password')
```
qwwqwwq: add `ng-model="old-password"` etc..

winterfell/webapp/src/js/modalChangePassword.js
```diff
@@ -0,0 +1,46 @@
...
+    return {
+        restrict: "A",
+        link: function(scope, elem, attrs) {
+            function displayError(msg) {
+              var errorMsg = $(elem).find('.error-msg');
+              errorMsg.text(msg);
+            }
+
+            scope.savePassword = function() {
+              var oldPwd = $(elem).find('input.old').val();
```
qwwqwwq: You don't need to use jQuery to find the value of this input, if you use `ng-model` it will be bound to the `scope` argument which is passed to this directive.

qwwqwwq:
Looks ok overall except for the one issue in the directive.

As a larger note this is why I'm not crazy about supporting password authentication in production. If we want to support it we need to figure out:

- changing passwords
- password resets including sending password reset emails
- storing hashed passwords in the database in production and the security risks involved

And probably a few more things I can't remember right now.. With oAuth2 we don't need to worry about any of these things. 

However users might prefer password authentication so we may not have a choice. I was hoping to find a library or service that helps manages all the above things but couldn't find anything that looked too good. We should be cautious about rewriting these base components that have already been done a million times over and probably have good open source implementations.

PR #88, opened 2016-10-10
Bunyan API Request Logger

ahsu1230:
Take a look at Bunyan: https://github.com/trentm/node-bunyan

Bunyan is a common nodeJs logger library, with lots of community support.
It does timestamps for you, logging levels, JSON objects well, and also has many flexible configurations to help you categorize different logging schemes.

In this change, you will see the following:
* Configure and setup Bunyan
* Have a main logger and a child logger (`logApi`) that acts as a middleware that will log incoming requests to the server and intercept outgoing responses so we can log what we send out.

To view logs, use the `followLogs.sh` script. It will open up the log file and prettify JSON objects for you. From this log file, you can choose to display only certain levels and above or choose categories (`widget_type` such as API).

For now, logging is only into a file. Soon, we can figure out a way to send things to an ElasticSearch without blocking on network calls.


winterfell/app.js
```diff
@@ -1,47 +1,60 @@
...
-}
-
-setupBiscuitKey(Constants.KEY_LOB_API);
-setupBiscuitKey(Constants.KEY_IDME_CLIENT_ID);
-setupBiscuitKey(Constants.KEY_IDME_SECRET_ID);
+// Initialize Logger
+var log = bunyan.createLogger({
+  name: 'vetafi_' + environment,
+  streams: [{
+    path: '../logs/app.js' // todo: these logs need to be moved into the same level
```
ahsu1230: this was taken out of winterfell because I was having trouble with pm2 and process.json. The server restarts everytime there is a change to any files. If the logging file is inside this overall folder, we will constantly be restarting.

`watch` is set to true, but even when trying to use `ignore_watch`, the server still kept restarting.

ahsu1230:
There's a bit of repeat on all controllers (`LogHelper.logApi`), but if I had a middleware that always called on every request... our logging will be littered when a client asks for the website .html, assets, etc. I'm not sure how to filter for our API endpoints and when a client requests assets.

winterfell/app.js
qwwqwwq: Hmm that is strange, I didn't think `pm2` had that functionality (restarting on change), is that something `gulp` might be doing?

qwwqwwq:
I tried out a library called `morgan` that does request logging, I will send out a revision with that soon, and we can maybe have it log through bunyan so the formatting is consistant.

winterfell/app.js
qwwqwwq: Also can we just set it to log to stream? pm2 will take care of writing the logs to a file for us / rotating files (see feature list here http://pm2.keymetrics.io/docs/usage/log-management/)  so we don't need to worry about setting a log output file here ..

winterfell/app.js
```diff
@@ -1,47 +1,60 @@
...
+  }]
+});
+app.log = log;                                            // Main logger
+app.logApi = app.log.child({widget_type: 'api'});         // Logger for API request/responses
+app.log.info("New logger created.");
```
qwwqwwq: Can we store the logger in a separate module and not attach it to the `app` object? This will allow us to log messages in modules that do not have easy access to the `app` instance (such as utilities/helper methods etc).

winterfell/app.js
```diff
@@ -1,47 +1,60 @@
...
 }
-
 loadIntoBuild(app, 'utils');
 loadIntoBuild(app, 'middlewares');
 loadIntoBuild(app, 'services');
 loadIntoBuild(app, 'controllers');
+Log.console("Node modules loaded.");
+
+// Initialize Biscuit
+app.set('secretsFile', require('./config/biscuit')[environment]);
```
qwwqwwq: Why did you move this here?

winterfell/controllers/authController.js
```diff
@@ -1,18 +1,17 @@
...
 var UserValues = require('./../models/userValues');
 var UserService = require('./../services/userService');
 
 module.exports = function (app) {
 
   // Endpoint for routing sign-up
-  app.get('/signup', function(req, res) {
-    console.log('[signup] request received for ' + JSON.stringify(req.body));
-    console.log('[signup] session for ' + JSON.stringify(req.session));
+  app.get('/signup', [Log.api], function(req, res) {
```
qwwqwwq: If you remove this manually added middlewares, I have a way to add a single middleware at app initialization and will log all the requests (using the `morgan` library).

winterfell/utils/logHelper.js
```diff
@@ -0,0 +1,46 @@
+var mung = require('express-mung');
```
qwwqwwq: we can use `morgan` instead of doing this

winterfell/app.js
ahsu1230: kk can do that.

winterfell/app.js
ahsu1230: just a reordering of operations. Wanted to log what initializing steps we passed in case it failed somewhere in the middle. That would mean logging would have to be first.
For biscuit, does the order matter?

On second thought it does since the app keys depend on biscuit to be set first. My bad, will swap the ordering so that biscuit comes before loading modules.

winterfell/controllers/authController.js
ahsu1230: I was just using a couple of different ways to log requests (including `morgan`, `bunyan-express`, `bunyan-request`, bunyan serializers) so we don't need to include this middleware everywhere. The problem with all of these is that with express, EVERYTHING gets logged. When you refresh the home page, logs for GET requests for assets, fonts, pictures, etc. all get logged - which becomes extremely noisy. 

I don't think those logs should get logged, only our APIs like `GET /user`. Which is why I did it the way I had it before (log only for every endpoint). I'm not sure if there's an easy way to filter out express only API requests.

ahsu1230: From a second look, a lot of companies usually have API endpoints in this format:
```
GET /api/user
POST /api/claim/:claimId
```

With this, you can do something like this:
```
app.all('/api/*', requireAuthentication, logStuff);
```

This is one way you can sort out api requests from the other network stuff.

qwwqwwq: I guess I'm not totally sure what the concern about having static assets requests logged is. I would rather have all requests for that reason that if someone requests a static asset and gets a 500 or a 404, we are still gonna wanna know about it because its something to fix. We can maybe mess around with the log levels (do DEBUG for 20x/30x on static assets), but the logger middleware should be on every end point.

ahsu1230: it gets noisy if all requests are on the same logging level.

ahsu1230:
For reference, when we view logs with bunyan's client, we can see the following:

`morgan` request logging looks like this:
```
[2016-10-20T02:39:54.992Z]  INFO: vetafi_local/30916 on packer-virtualbox-iso-1469913211: (widget_type=api)
    ::ffff:10.0.2.2 - - [Thu, 20 Oct 2016 02:39:54 GMT] "GET /api/user HTTP/1.1" 304 - "http://localhost:3999/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.143 Safari/537.36"
```

using bunyan request serializers logging looks like this:
```
[2016-10-20T04:48:23.656Z]  INFO: vetafi_test/4475 on packer-virtualbox-iso-1469913211:  (widget_type=api, req.params={}, req.query={}, req.requestId=b4294e2a-7238-4e93-99e5-21f7171c36e3)
    GET /api/user HTTP/1.1
    host: 127.0.0.1:41831
    accept-encoding: gzip, deflate
    cookie: connect.sid=s%3AwEgorHObTbkaFjNWMy9bxiNWfq-Qi99f.ecujI5nscQW7HLh9xplWkhxUYyqAmOsR4tpt%2FJ9pIJM
    user-agent: node-superagent/1.8.3
    connection: close
    
    {}
[2016-10-20T04:48:23.660Z]  INFO: vetafi_test/4475 on packer-virtualbox-iso-1469913211:  (widget_type=api)
    resp: {
      "status": 200,
      "method": "GET",
      "url": "/api/user",
      "requestId": "b4294e2a-7238-4e93-99e5-21f7171c36e3"
    }
    --
    body: {
      "user": {
        "firstname": "Sir",
        "middlename": "Moose",
        "lastname": "Alot",
        "email": "sirmoosealot@test.com",
        "externalId": "be61c9ac-a340-43b0-8d60-d94925f1aa0e"
      }
    }
```

I think this works a lot better with bunyan because Morgan converts everything into a string, but bunyan is more object oriented so bunyan displays it neatly and allows better filtering through its object oriented conditionals.
Also with serializers, it'll just be part of the request middlewares so no need to include the Logger everywhere ^^.
`express-mung` is still needed because morgan only takes care of requests not responses.

ahsu1230:
@qwwqwwq I made a few more changes, could I get another look please? 😄 
It's a lot better this time. 
* The logging is automatic for every request so we don't have to inject it everywhere. 
* It separates API requests from asset requests by filtering `/api/` for endpoints. api requests are info level and assets are debug.
* It also logs responses
* Logs are more bunyan friendly so they are easier to read.

PR #89, opened 2016-10-13
Use autocomplete Attribute for Form Elements

qwwqwwq:
The `autocomplete` attribute gives us control over the way browsers will fill in our forms (https://html.spec.whatwg.org/multipage/forms.html).

In this revision I have modified the formly library to allow passing an `autocomplete` parameter, and updated the form template to provide it for elements where it makes sense.

I plan to submit my change as a pull request to the formly js project, but I am just modifying the library code in-place for now. I like that we have our dependencies checked in, which allows us to modify our libraries like this when needed.

### DIFF
--- a/winterfell/forms/VBA-21-0966-ARE.json
+++ b/winterfell/forms/VBA-21-0966-ARE.json
@@ -49,7 +49,8 @@
         "email"
       ],
       "templateOptions": {
-        "label": "Your Email"
+        "label": "Your Email",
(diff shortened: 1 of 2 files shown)


ahsu1230:
lgtm, nice thought.
I do want to caution that if we want to upgrade the formly library, we would have to rebase these changes on top of the upgraded library. If we had this logic in a wrapper over the library, it would be ideal, but should be fine for now.

PR #90, opened 2016-10-17
Adds Email Service Backed By SES

qwwqwwq:


winterfell/services/mailingService.js
```diff
@@ -1,10 +1,11 @@
 var lob = require('Lob');
```
ahsu1230: these injections can be alphabetized

PR #91, opened 2016-10-19
Adds eslint

qwwqwwq:
eslint is a popular linting and autoformatting library, here we add configurations for eslint to the repo, we can eventually use this library to do thinks like enforce import order, etc.


qwwqwwq:
This should be ready to land now, it will show linting errors when we run gulp build.

winterfell/package.json
```diff
@@ -57,5 +57,11 @@
     "underscore": "^1.8.3",
     "uuid": "^2.0.2",
     "yargs": "^4.7.1"
+  },
+  "devDependencies": {
```
ahsu1230: nice

winterfell/webapp/.eslintignore
```diff
@@ -0,0 +1,2 @@
+src/libs/**
```
ahsu1230: isn't this already included in the `eslintignore` file above?

winterfell/webapp/gulpfile.js
```diff
@@ -129,6 +138,6 @@ gulp.task('watch', ['build', 'initBrowserSync'], function () {
   gulp.watch('src/**/*.jade', ['jade']);
```
ahsu1230: you probably want to run the `lint` task after you recompile javascript files from a refresh. (line 136)

winterfell/webapp/.eslintignore
qwwqwwq: supposedly the top level eslintignore file is supposed to cascade down but in practice i wasnt seeing this happen, will dig in more

winterfell/webapp/gulpfile.js
qwwqwwq: well all it does right now is spam a bunch of stuff out to the console, which would be annoying during rapid iteration

i think it should just be in the main build, and we can have it fail the build task if there are lint errors, which will force us to keep our code clean

i didnt turn that feature on yet, because we still need to clean up all the code

qwwqwwq:
Ok I addressed the redundancy in the `.eslintignore` files, I didnt change the gulpfile though see my comment for my reasoning there. Any editor you use should be able to parse these files and offer you lint warnings now, at least it works for me in intellij

qwwqwwq:
ping

ahsu1230:
lgtm

PR #93, opened 2016-10-19
Fix Select Forms page, forms not showing up

ahsu1230:
This change does the following:
- For formly-form objects, anything that is custom defined by Vetafi can be in a smaller object `vfi`.

For example, this form "Intent to File" has its own short title and description and every claim is required to file it. Hence, the `required` field.
- We should be displaying ALL forms available. Then, the user will select which forms they would like to fill out. So, the angular UI will display all forms in `formTemplateService`. Using information from `$scope.claimForms`, which is forms in that claim, we can see what state the form is in (i.e. required but completed?)


qwwqwwq:
A couple questions:

- If a claim is allowed to include ANY set of forms, what is the definition of a claim? Does a set of forms being in a claim mean they are just mailed together, and nothing else? I imagined that when a user selects to file an "intent to file" claim, they should only see one form option, because that is the only applicable form for them.
- The required attribute is kind of confusing for me. If a claim is meant to achieve a specific goal, it makes sense to just show the user all the forms they need to complete to get to their goal, and the fact that they are required is implied. We wouldn't ask the user to do it if it wasn't required, because the point here is to do the absolute minimum work.

Finally we really should land https://github.com/bayeshack2016/vetafi/pull/86 before we start changing things here, because that revision touches this stuff a lot.

ahsu1230:
Ah okay.

I've been operating under the premise that a claim is simply a way to connect a user to a list of forms that get mailed/emailed. A claim should be very generic, I don't think we should have "types" of claim. An "Intent to file" claim isn't its own type - it is just a collection of "Intent to file" forms (in this case, one form) that gets sent to addresses and emails.

When selecting forms to fill, I think it's going to be very challenging to predict what forms are going to be important to a user. From filling out an Intent to File, and a couple of general info forms, how would we know that this veteran needs help/aid with his home or lifestyle? How do we know he/she wants to include sexual assault trauma into their claim? What about tending to your spouse or child's needs? If you want to include PTSD into the claim, the path may not have a step-by-step decision tree.

So what I'm saying is the decision tree of only showing what's required works well for form answers, but it's not as clear for selecting forms. TurboTax has pages dedicated to giving you options with what you might be interested in like so:

tt2014_4
turbotax-2015-review-business

The first picture, you'll see this at the beginning. You click on a category then start that. One you're done filling out the first category, you'll be brought to something similar to the second picture.

The second picture (which is for business users, but there's a similar screen for consumers) shows when the user is mostly done filling out the essential forms, then is given the option to go back and edit the categories they just finished or continue to other categories/sections. This method also saves our ass because we show ALL options available to the user - we aren't hiding anything so can't be liable for misinformation.

What we can do (just like what we did in our MVP) is highlight certain forms that are required or recommended (still optional). So even though we're showing a lot of options, we don't have to hide things, all we have to do is make important options more visible to our users. And there are a million things we can do to do that such as coloring, icons, etc. We can even include a filter/search bar where you can search for forms that you'll be interested in, if needed.

So I think when selecting forms, we should give as many options as possible. From there, whatever forms that end up being selected/filled will be part of the current claim. For our modeling, a claim is generic and only needs to know that there's an owner, there are forms, whether it is submitted, and where it is submitted to.

What do you think?

qwwqwwq:
Awesome research and points, thanks for finding those screenshots from turbotax those are definitely great examples.

I don't think we are that far off from each other actually. I agree it is going to be difficult to predict what the user wants to accomplish (even the user themselves may not know what they want to accomplish as they start out), so I like the idea of offering them all the options and allowing them to pick and choose.

What I was thinking is that the forms themselves are pretty low level, and are more or less meaningless to the user. It may take some arbitrary set of forms to accomplish a high level goal, i.e. getting mental health treatment might require applying for health insurance and filling out a PTSD form. In my mind the claims would be the meaningful entities to the user. 

The user would choose a set of claims that meet the goals they want to accomplish, and then within each claim they would have to answer a bunch of questions that just so happen to correspond to forms. At the end of the day, the users will just know "I filed a claim for disability and a claim for mental health treatment" instead of "I filed forms A, B, and C". Having the higher level abstraction of claims is a tool that allows us to hide unnecessary complexity from the user.

ahsu1230:
Ohhh I see. That's a pretty neat idea. We definitely eventually need some kind of categorization.
But, we should still send one claim and all of its forms together.

Similar to how TurboTax does things, there's a general tax info flow (where you list your income, etc.) then there are other flows for specific categories (mortgages, auto, insurances, charities, etc.). Each flow has its own decision tree and form questions. The second picture in my previous comment shows this. At the end of each flow, the user gets to see more categories and if they are done, TurboTax files all of the information together. 

We can eventually have a flow for PTSD, and another flow for health treatment, etc. In each of these flows, a user picks out certain forms and then answers certain questions. But, at the end of the day, once the user has felt that he/she has completed all of the necessary flows, we stitch the package together and file the claim as one letter. I think it gets super confusing if we end up sending one letter for a PTSD claim, and then another for health treatment. Not to mention, there will probably be dropoffs (after filing a PTSD claim, i'm less likely to file a health treatment one).

Database model wise, I think having a `claim` just care about a list of `forms` regardless of what they are makes our lives a lot easier.

So for MVP, since we only have one form, we probably don't need any categories, so let's just have users select forms. Once we finish supporting all general forms and move onto more specific forms, we can start thinking of categories and other flows. How's that?

qwwqwwq:
Ok I see your point now. In the steady state we will probably want to only do one submission per user.

So let's change select forms to list all forms. There may also need to be some changes to how the api works for the select forms page, as it will be creating new Form documents instead of updating them.

ahsu1230:
@qwwqwwq this is ready for another review

winterfell/webapp/src/js/claimSelectFormsCtrl.js
```diff
@@ -1,25 +1,29 @@
-var app = angular.module('vetafiApp');
-app.controller('claimSelectFormsCtrl', ['$scope', 'claimService', 'formTemplateService', '$stateParams', 'forms', 'net',
-  '$state',
-  function($scope, claimService, formTemplateService, $stateParams, forms, net, $state) {
+var app = angular.module('vetafiApp');``
```
qwwqwwq: I think there are some extra backticks at the end of this line?

winterfell/webapp/src/templates/claimSelectForms.pug
```diff
@@ -9,15 +9,15 @@
         uib-progressbar.progressbar(value="numRequiredCompleted", max="numRequiredForms", type="{{numRequiredCompleted == numRequiredForms ? 'success' : 'info'}}")
           span {{numRequiredCompleted}}/{{numRequiredForms}}
 
-      div.form-wrapper(ng-repeat="(formId, form) in forms")
+      div.form-wrapper(ng-repeat="(formId, form) in allForms", ng-init="myForm = myForms[formId]")
```
qwwqwwq: what does ng-init do?

winterfell/webapp/src/templates/claimSelectForms.pug
```diff
@@ -9,15 +9,15 @@
...
-            h5.title(ng-class="{'required': formTemplates[form.key].required}") {{form.id}}
-            h5.warning.required(ng-if="form.state == formTemplates[form.key].required") &nbsp;(required)
-            h5.subtitle {{formTemplates[form.key].unofficialTitle}}
-          div.form-summary {{formTemplates[form.key].unofficialDescription}}
+            h5.title(ng-class="{'required': form.vfi.required}") {{formId}}
+            h5.warning.required(ng-if="form.vfi.required && !(myForm && myForm.answered > 0 && myForm.answered == myForm.answerable)") &nbsp;(required)
+            h5.subtitle {{form.vfi.title}}
+          div.form-summary {{form.vfi.summary}}
         div.form-buttons
           button.edit-btn(ui-sref="root.form({claimId: claimId, formId: formId})")
```
qwwqwwq: does this link still work? Shouldn't formId now be the name of the form and not the mongo identifier?

winterfell/webapp/src/templates/claimSelectForms.pug
ahsu1230: it allows you to set a variable.
Instead of having `myForms[formId]` everywhere, now I can set it to `myForm` and use that variable instead.

winterfell/webapp/src/js/claimSelectFormsCtrl.js
ahsu1230: woops

winterfell/webapp/src/templates/claimSelectForms.pug
ahsu1230: Yes `formId` is the name of the form (look at line 17 where `formId` denotes the name of the form).

qwwqwwq:
Ok LGTM

PR #94, opened 2016-11-01
Log To Stream Instead of File By Default

qwwqwwq:


winterfell/middlewares/log.js
```diff
@@ -9,7 +9,7 @@ function Log(app) {
   mainLog = bunyan.createLogger({
     name: 'vetafi_' + app.environment,
     streams: [{
-      path: '../logs/app.js', // todo: these logs need to be moved into the same level
+      stream: process.stdout,
```
ahsu1230: cool we could do it this way. However, when it comes to looking at json objects in a clearer fashion, it's nicer to use the bunyan client. Like so: 
```
tail -F ../logs/app.js | ./node_modules/.bin/bunyan
```

Perhaps we can do the same with pm2?

ahsu1230:
There's a file called `followLogs.sh`.
Maybe we could replace it to look like this:
```
pm2 logs Express | ./node_modules/.bin/bunyan
```

Using the bunyan client will help pretty-print the json objects. Not sure if the above command would work, but it might

PR #95, opened 2016-11-01
Update Jade -> Pug

qwwqwwq:
Jade is not called jade anymore, its called pug, we should fix this now before it becomes confusing in the future.


ahsu1230:
thanks!! lgtm

PR #96, opened 2016-11-03
Render Login and Signup Pages Server Side

qwwqwwq:
This renders the login and signup pages server side in response to request, allowing us to embed a csrf token, as opposed to using static pages as we did previously.


winterfell/views/login.pug
```diff
@@ -8,9 +8,10 @@ block emailText
   div.txt-container Login with your email
 
 block inputs
-  form
-    input.email(placeholder='Email Address', type='text')
-    input.pwd(placeholder='Password', type='password')
+  form(action='/api/auth/login' method='post')
```
ahsu1230: Missing onsubmit validateForm

winterfell/controllers/authController.js
```diff
@@ -57,13 +69,13 @@ module.exports = function (app) {
...
                 req.session.userId = req.user._id;
-                return res.status(http.OK).send({redirect: '/'});
+                return res.redirect('/');
               });
             }
           );
         });
       } else { // User does exist!
-          res.status(http.BAD_REQUEST).send({error: httpErrors.USER_EXISTS});
+          res.redirect('/signup?error=EUSERNAMETAKEN');
```
ahsu1230: When we redirect to signup or login with an error message we should handle that on the front end JavaScript side. ,Anne look for the Parma and if it's there call the displayError message?

winterfell/controllers/authController.js
```diff
@@ -76,23 +88,10 @@ module.exports = function (app) {
       var extUserId = req.user.externalId;
       res.status(http.OK).send({userId: extUserId, redirect: '/'});
     } else {
-      res.status(http.UNAUTHORIZED);
+      res.redirect('/login?error=EAUTHFAILED');
```
ahsu1230: Same here

winterfell/app.js
```diff
@@ -55,6 +55,9 @@ Log.console("New logger created.");
 // Connect to a mongodb server using mongoose
 require('./config/mongoose')(environment);
 
+// Set templating engine
+app.set('view engine', 'pug');
```
ahsu1230: Oh neat

winterfell/views/login.pug
qwwqwwq: The previous code did not do any validation for login AFAICT, what validation would need to be done?

winterfell/controllers/authController.js
qwwqwwq: fixed

winterfell/views/login.pug
ahsu1230: oh really? maybe at least make sure the input is a valid email?
If not, nvm

winterfell/utils/constants.js
```diff
@@ -22,3 +22,12 @@ module.exports.biscuitKeys = {
   idmeClientId: 'id-me-client-id',
   idmeSecretId: 'id-me-client-secret'
 };
+
+module.exports.ERROR_CODES = {
```
ahsu1230: ERROR_CODES constant keys are good for server-side. But I think the message that it is associated with it, should be on the front end.
So I think the server response could send a `eusernametaken` key back.
But on the front end, it is more proper to take that key to get a message.

ahsu1230:
LGTM. Though I do prefer to have the error message on the front end as opposed to the backend.

winterfell/views/login.pug
qwwqwwq: my opinions on doing email validation are: https://davidcel.is/posts/stop-validating-email-addresses-with-regex/

winterfell/utils/constants.js
qwwqwwq: I would argue that since the error originates from the backend, so it makes sense to have its message stored there too. This allows some nice decoupling of the front and back ends. It would be brittle to expect the client to maintain a hashmap of error codes that must stay in sync with the set of error codes the server may produce.

PR #97, opened 2016-11-03
Adds Express Best Security Practices

qwwqwwq:
- Use real session secret, secure session cookie.
- Use helmet middleware.
- Use global rate limiting per user and IP.
- Use CSRF protection for all angular and non-angular parts of the site.


qwwqwwq:
ping

winterfell/middlewares/expressLimiter.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+function addLimiter(app) {
+  if (app.environment === Constants.environment.PROD) {
+    var limiter = expressLimiter(app, client);
+
+    limiter({
+      path: '/',
+      method: 'all',
+      lookup: ['connection.remoteAddress'],
+      // 360 request per hour per address for all pages
```
ahsu1230: is 360/hr standard? I feel like it should be within a smaller window.
Attacks usually happen when a lot more requests are sent within a few minutes.

winterfell/middlewares/log.js
```diff
@@ -42,6 +42,22 @@ module.exports.console = function(msg) {
   console.log(msg);
 };
 
+module.exports.info = function(msg) {
```
ahsu1230: nice I was thinking this too

ahsu1230: I know Android shortens it to
```
Log.i(...)
Log.d(...)
Log.w(...)
```
if we want to shorten things

qwwqwwq: My personal style is towards more verbose but easier to read for people new to the code, not sure what the node convention is though

PR #98, opened 2016-11-05
Fix Bug in Claim Submit Flow

qwwqwwq:
- Fix api endpoint
- Fix form completion logic to consider signature as a question.


winterfell/services/claimService.js
```diff
@@ -92,16 +95,16 @@ module.exports.findIncompleteClaimOrCreate = function(userId, forms, callback) {
   });
 };
 
-module.exports.addForm = function(claimId, file, callbacks) {
+module.exports.addForm = function (claimId, file, callbacks) {
```
ahsu1230: ok I assume a space after function is standard now? haha

winterfell/webapp/src/js/claimSelectFormsCtrl.js
```diff
@@ -9,6 +9,7 @@ app.controller('claimSelectFormsCtrl', ['$scope', 'claimService', 'formTemplateS
     });
 
     $scope.numRequiredCompleted = _.sum(_.map(forms, function (form) {
+      console.log(form);
```
ahsu1230: still need this?

ahsu1230:
lgtm, but looks like there's a merge conflict!

winterfell/services/claimService.js
qwwqwwq: I have no idea what is standard and what is not in javascript TBH, once we turn on the linter it should all be taken care of and standardized.

winterfell/webapp/src/js/claimSelectFormsCtrl.js
qwwqwwq: Nope, these statements will also be tagged by the linter for removal once we turn it on.

PR #99, opened 2016-11-11
Configures Packer to run Amazon-EBS Builder

qwwqwwq:


ahsu1230:
is there a resource that you followed to do this? Mind linking it?

qwwqwwq:
Yeah! Mostly used the docs here: https://www.packer.io/docs/builders/amazon-ebs.html

Also we use this process at work so I've done it a bunch of times before

config/packer/vetafi-ubuntu-14.04.4-server-amd64.json
```diff
@@ -109,6 +127,11 @@
                 "provider": "virtualbox",
                 "version": "0.0.4"
             }
-        }]
+        },
+        {
+            "type": "atlas",
+            "artifact": "qwwqwwq/nodejs-mongodb-redis",
```
ahsu1230: I've noticed that there are a lot of personal folders in a lot of these paths. Can we make them a globally named path of sorts? This way, we can more confident things work/don't work based on personal local environments.

config/packer/vetafi-ubuntu-14.04.4-server-amd64.json
```diff
@@ -1,6 +1,6 @@
 {
     "push": {
-      "name": "",
+      "name": "qwwqwwq/nodejs-mongodb-redis",
```
ahsu1230: same here

config/packer/vetafi-ubuntu-14.04.4-server-amd64.json
qwwqwwq: This isn't a personal folder, this is the name of our image on atlas. When I signed up for atlas I used the username qwwqwwq.

PR #108, opened 2016-11-15
Every claim saves where copies were sent to

ahsu1230:
Copies of claims can be sent to a list of emails and a list of addresses.

For now, we assume the user's emails.
For addresses, we assume the VA and the user's address (which they should've filled out).

There may need to be one more change (after this PR) to fix all user_values mapping to follow the standard AddressSchema.
We want to use `street1` instead of `line1` from now on and `postal` instead of `zip`.


qwwqwwq:
LGTM

qwwqwwq:
for user values, yeah we can use the standard address terms i think

PR #109, opened 2016-11-16
Fix CSRF In Tests

qwwqwwq:
I failed to noticed that the addition of the csurf middleware had broken all the tests, this fixes that br providing the token when needed (it is needed for any test that does an API POST or DEL)


ahsu1230:
lgtm

PR #111, opened 2016-11-17
relink idme for sign in pages

ahsu1230:
Relink Id.Me redirect url from button. Remember that the url depends on local or prod environment as the redirectUri for Id.Me is different.

As a bonus, move the link from signup -> login or login -> signup to the top layer (instead of in the bottom layer)

### DIFF
--- a/winterfell/controllers/authController.js
+++ b/winterfell/controllers/authController.js
@@ -9,6 +9,9 @@ var UserValues = require('./../models/userValues');
 var UserService = require('./../services/userService');
 
 module.exports = function (app) {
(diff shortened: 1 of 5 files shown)


ahsu1230:
yes, the original logic was that it will use whatever email you used for Id.Me and check if it's a new email.
If it's a new email and the social token is not tied to any user, we can create a new user!
Otherwise, if the email exists and the social token matches, we sign in as that existing user.

I don't have good error states/redirects though (i.e. email exists, but the social tokens don't), but oh well

ahsu1230:
Also, Facebook hasn't been done yet. Though I made the Id.Me generic enough where it can re-use a lot of the facebook logic. Should we include Facebook into MVP or nah (and just take out the UI)?

qwwqwwq:
Oh ok, at this point, id say we should take it out if it doesn't work and is going to be a lot more effort, might cause confusion when demoing if its there but not functional.

PR #114, opened 2016-12-01
fix eslint errors on webapp side

ahsu1230:
Es-lint produces a lot of errors and warnings. This is to remove mostly unused variables.

i am still having problems with removing `console.log` warnings and 
```
warning  This module could be parsed as a valid script
```

### DIFF
--- a/winterfell/webapp/.eslintrc
+++ b/winterfell/webapp/.eslintrc
@@ -1,5 +1,4 @@
 {
-    "extends": "eslint:recommended",
     "env": {
         "browser": true,
         "node": false
@@ -8,5 +7,9 @@
         "_": true,
         "$": true,
         "angular": true
-    }
+    },
+    "rules": {
+      "no-console": "off"
+    },
+    "extends": "eslint:recommended"
 }
--- a/winterfell/webapp/src/styles/header.styl
+++ b/winterfell/webapp/src/styles/header.styl
@@ -108,13 +108,7 @@
       cursor: pointer
 
 #banner
-  background: repeating-linear-gradient(
-    45deg,
-    colorPrimaryBlue,
-    colorPrimaryBlue 10px,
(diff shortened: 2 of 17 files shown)


winterfell/webapp/.eslintrc
```diff
@@ -8,5 +7,9 @@
         "_": true,
         "$": true,
         "angular": true
-    }
+    },
+    "rules": {
+      "no-console": "off"
```
ahsu1230: I was under the impression that this is enough, but it's still not ignoring console.log in our code :/

winterfell/webapp/src/styles/header.styl
```diff
@@ -108,13 +108,7 @@
...
 
 #banner
-  background: repeating-linear-gradient(
-    45deg,
-    colorPrimaryBlue,
-    colorPrimaryBlue 10px,
-    colorPrimaryDarkestBlue 10px,
-    colorPrimaryDarkestBlue 20px
-  );
+  background: colorDarkGray
```
ahsu1230: sorry I think the gradient is way too sharp and distracting. 😖 Turned it into a more neutral color and the text will still be more viewable.

qwwqwwq:
Yeah I couldnt figure out what `warning  This module could be parsed as a valid script` meant either..

winterfell/webapp/.eslintrc
qwwqwwq: It might be picking it up from the parent config.. oh well

winterfell/webapp/src/styles/header.styl
qwwqwwq: Thats fine, youre much better at design than I am, I trust your opinion

qwwqwwq:
LGTM

ahsu1230:
Thanks for the review!

PR #117, opened 2016-12-01
ToS page with words

ahsu1230:
A separate page for Terms of Service with real words copied mostly from TurboTax
https://turbotax.intuit.com/corp/terms-of-service.jsp (obviously Vetafi replacing Intuit).

The agreement modal also uses these paragraphs. And then some style alterations between the page form and modal form.


qwwqwwq:
LGTM

PR #118, opened 2016-12-02
FAQ page with real information and links

ahsu1230:
the `tags` variable you see in every object is used to help with searching.

There is a search bar which you can type keywords to find things you want. The search bar simply looks at words in the title, content and tags. So if we don't mention the word in the description, we can catch it through the use of these tags.

### DIFF
--- a/winterfell/webapp/src/js/faqCtrl.js
+++ b/winterfell/webapp/src/js/faqCtrl.js
@@ -1,20 +1,70 @@
 var app = angular.module('vetafiApp');
+
+app.filter('unsafe', function($sce) {
+  return $sce.trustAsHtml; }
+);
+
 app.controller('faqCtrl', ['$scope', function($scope) {
   $scope.sections = [
     {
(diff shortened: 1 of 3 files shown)


winterfell/webapp/src/js/faqCtrl.js
```diff
@@ -1,20 +1,70 @@
 var app = angular.module('vetafiApp');
+
+app.filter('unsafe', function($sce) {
```
qwwqwwq: What is this about?

ahsu1230: this is used in `faq.pug`
```
div.section-content(ng-show="section.html", ng-bind-html="section.html | unsafe")
```
It's for injecting html content, but you can only do it with the $sce module (well one of the easiest way). Otherwise, Angular won't do it and will give back an error.
https://docs.angularjs.org/api/ng/service/$sce

PR #120, opened 2016-12-04
slightly alter home page

ahsu1230:
After looking at TurboTax's home page, I was inspired for a bit to change our home page to have two things

1) We need a header of sorts to explain what we do in a very short way.
TurboTax says: "Taxes done smarter, together"

2) The buttons to link other pages have looked pretty ugly to me for a while. I've changed them one more time to make it more interesting and Silicon-Valley like. I hope you like it!


qwwqwwq:
Looks awesome!

PR #121, opened 2016-12-04
about us page

ahsu1230:
An About Us page for going live and who this product is by.


winterfell/webapp/src/templates/aboutus.pug
```diff
@@ -0,0 +1,37 @@
...
+      h3 Our Founders
+      div.card-wrapper.container-fluid
+        div.row
+          div.profile-wrapper.col-sm-3
+            div.profile-back
+              div.profile-pic.jeff
+          div.content-wrapper.col-sm-9
+            h4 Jeffrey Quinn
+            p
+              | Jeff is currently a Data Engineer at the health startup company, #[a(href="https://www.nuna.com/") Nuna].
```
ahsu1230: Let me know if you want a different description. I just put stuff down.

ahsu1230:
let me know if you want to use a different profile pic as well.

qwwqwwq:
This seems like a bit of an odd thing to have on the user-facing site from my perspective, what is the end goal here?

ahsu1230:
Many websites have this - especially non-profit services. It simply builds rapport and trust with the community of people who want to use this site. One of the best ways to differentiate between a scamming site is to provide identities. 
https://www.rocketspark.com/blog/why-about-us-page-so-important/

This page isn't needed now, but it was a link in the footer so I figured I would just include it. We can always remove the link in the footer if we end up not wanting this page at all.
Thoughts? I know it can be a bit weird putting our identity out there.

qwwqwwq:
Ok I agree it might be a good idea to have this, not too worried about putting my identity out there, just don't want to mess up the vetafi brand

PR #127, opened 2016-12-06
hashed passwords in db

ahsu1230:
use `bcrypt` to hash passwords before saving them to db.
Then use `bcrypt` to compare passwords.

This mostly affects tests as we assumed the original password with the crsf tokens.

production wise: PREVIOUS ACCOUNTS WONT WORK ANYMORE.
I think it's okay because we have no users?

### DIFF
--- a/winterfell/services/authService.js
+++ b/winterfell/services/authService.js
@@ -1,21 +1,23 @@
 var _ = require('lodash');
+var bcrypt = require('bcryptjs');
 var SocialUser = require('../utils/socialUser');
 var User = require('../models/user');
 var UserValues = require('../models/userValues');
 var UserService = require('../services/userService');
+
+var salt = bcrypt.genSaltSync(10);
 
 function AuthService (app) {
     this.app = app;
 };
 
 module.exports = AuthService;
+module.exports.generatePassword = function(pwd) {
+  return bcrypt.hashSync(pwd, salt);
+};
+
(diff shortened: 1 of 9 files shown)


qwwqwwq:
I wouldnt worry about previous accounts not working, pretty clearly there is that huge beta mode banner on the site right now and its been up and down this whole time..

I started looking at the best practices for authentication used here: https://github.com/robert52/express-starter, its a pretty useful guide. Specifically I think we should emulate the way he sets its up so that the hashed password can never actually leave the database layer..

I started implementing his approaches a few months ago but never went through with it.

winterfell/services/authService.js
```diff
@@ -1,21 +1,23 @@
 var _ = require('lodash');
+var bcrypt = require('bcryptjs');
 var SocialUser = require('../utils/socialUser');
 var User = require('../models/user');
 var UserValues = require('../models/userValues');
 var UserService = require('../services/userService');
 
+var salt = bcrypt.genSaltSync(10);
```
qwwqwwq: Each user should have their own salt, otherwise it defeats the purpose of the salt, cryptographically speaking

ahsu1230:
I see, so I would need to do the following.
1) Every user would have to save their own passwordSalt
2) When returning the user object to services/controllers, password and passwordSalt should not be part of it.

PR #130, opened 2017-01-05
fix many small bugs

ahsu1230:
Fixes the following bugs:

* Continue Claim on home page
* Login redirecting
* Login Page (links should be their own line block, instead of inline)
* Logout link
* Select Forms page: Fix icon and font when a required form is complete
* Claim Confirm page: name not showing up
* View Claim page: emails and addresses claim was sent to
* View Claim page: name and pdf download

### DIFF
--- a/winterfell/controllers/authController.js
+++ b/winterfell/controllers/authController.js
@@ -105,8 +105,7 @@ module.exports = function (app) {
       if (req.user) {
         req.session.key = req.body.email;
         req.session.userId = req.user._id;
-        var extUserId = req.user.externalId;
-        res.status(http.MOVED_TEMPORARILY).send({userId: extUserId, redirect: '/'});
+        res.redirect('/');
       }
   });
 
--- a/winterfell/webapp/src/js/net.js
+++ b/winterfell/webapp/src/js/net.js
@@ -24,20 +24,6 @@ app.factory('net', ['xhrEnv', '$http', function(xhrEnv, $http) {
   };
 
   return {
-    // Auth
-    login: function (email, password) {
-      var data = {
-        email: email,
-        password: password
-      };
-      return httpPost("/auth/login", data);
-    },
-    logout: function() {
-      return httpGet("/auth/logout");
-    },
-    signup: function(userData) {
-      return httpPost("/auth/signup", userData);
-    },
     touchSession: function() {
       return httpGet("/session/touch");
     },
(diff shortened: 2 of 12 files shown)


winterfell/controllers/authController.js
```diff
@@ -105,8 +105,7 @@ module.exports = function (app) {
       if (req.user) {
         req.session.key = req.body.email;
         req.session.userId = req.user._id;
-        var extUserId = req.user.externalId;
-        res.status(http.MOVED_TEMPORARILY).send({userId: extUserId, redirect: '/'});
+        res.redirect('/');
```
ahsu1230: frontend used to handle redirect. Now server does.

winterfell/webapp/src/js/net.js
```diff
@@ -24,20 +24,6 @@ app.factory('net', ['xhrEnv', '$http', function(xhrEnv, $http) {
   };
 
   return {
-    // Auth
```
ahsu1230: unused

winterfell/webapp/src/js/profileCtrl.js
```diff
@@ -58,15 +58,6 @@ app.controller('profileCtrl', ['$scope', '$location', '$window', 'Profile', 'cla
       $window.location.href = net.getAuthIdMeUrl();
     };
 
-    $scope.clickLogout = function() {
```
ahsu1230: better to use `<a>` link instead of using `$location`

winterfell/webapp/src/templates/claimConfirm.pug
```diff
@@ -17,7 +17,7 @@
       h4 General
       div.basic-row
         span Name:
-        span {{user.firstName}} {{user.lastName}}
+        span {{user.firstname}} {{user.lastname}}
```
ahsu1230: typo not camelCase

winterfell/webapp/src/templates/claimView.pug
```diff
@@ -42,7 +42,7 @@
         div.form-header
           div.form-id {{form.key}}
           div.form-title {{form.name}}
-        a(ng-href="/api/claim/{{claimId}}/form/{{form.key}}/pdf")
+        a(ng-href="/claim/{{claimId}}/form/{{form.key}}/pdf")
```
ahsu1230: no `/api`

winterfell/controllers/authController.js
qwwqwwq: good

winterfell/webapp/src/js/profileCtrl.js
qwwqwwq: agree

PR #133, opened 2017-01-07
allow overwriting of uservalues

ahsu1230:
the front end is repeatedly sending data that is updating the uservalues.
We need to remove this overwriting check


ahsu1230:
https://github.com/bayeshack2016/vetafi/issues/132

qwwqwwq:
Hmm this makes sense now. I guess if the person changes something in a form, obviously 99% of the time we will want to reflect that change back to the user values. LGTM

PR #134, opened 2017-01-19
claim submit error should route to submit-error page

ahsu1230:
When there is an error on the `submit` endpoint, route to the submit error page.

* a small bug change with name fields
* a few style changes on home screen

### DIFF
--- a/winterfell/webapp/src/js/app.js
+++ b/winterfell/webapp/src/js/app.js
@@ -171,7 +171,7 @@ app.config(['$stateProvider', '$urlRouterProvider', function ($stateProvider, $u
 
   $stateProvider.state({
     name: 'root.claimsubmit',
-    url: '/claim/{claimId}/submit',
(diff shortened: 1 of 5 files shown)


winterfell/webapp/src/js/app.js
```diff
@@ -171,7 +171,7 @@ app.config(['$stateProvider', '$urlRouterProvider', function ($stateProvider, $u
 
   $stateProvider.state({
     name: 'root.claimsubmit',
-    url: '/claim/{claimId}/submit',
+    url: '/claim/{claimId}/submit?error',
```
qwwqwwq: Do we need the null query param to prevent an error in the client JavaScript?

ahsu1230: on the submit page, we check if the error is truthy (not undefined or null)
so if there's no query param for `error`, we're ok. There might be a cleaner way to do this though

PR #136, opened 2017-01-21
Show Loading Animation When Busy

qwwqwwq:
- Adds loading SVG to static assets
- Adds hidden fixed div to root template with the loading svg.
- Adds HTTP interceptor that detects AJAX requests and starts a timer. When the timer completes, if there are still requests pending, the loading animation is revealed. When all requests are complete, the loading animation is hidden.
- On the forms page, reduce AJAX requests that are issued as part of autosaving by saving the data from the last autosave, and not issuing a request if nothing has changed. Without doing this the animation could blink on in an annoying way.
- Also fixed unrelated bug with misnamed variable.


winterfell/webapp/src/js/formCtrl.js
```diff
@@ -44,13 +44,15 @@ app.controller('formCtrl', ['$scope', '$filter', '$rootScope', 'formTemplateServ
...
-          console.log(response);
-        });
+        save(true)
       };
 
-      function save() {
-        return net.saveForm($stateParams.claimId, $stateParams.formId, $scope.model);
+      var lastParams = null;
+      function save(force) {
+        if (lastParams == null || !_.isEqual(lastParams, $scope.model) || force === true) {
```
ahsu1230: oh nice

winterfell/webapp/src/styles/main.styl
```diff
@@ -87,3 +87,23 @@ body
...
 		line-height: lead_lineHeight
+
+#busy-overlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100vw;
+  height: 100vh;
+  z-index: 1000;
+  background-color: grey;
```
ahsu1230: use our custom `colorWarmDarkGray`?

PR #142, opened 2017-01-27
Update VBA-021-0966 Form to Latest Version

qwwqwwq:
- Update questions and pdf locators.
- Force country to be two letters (was causing issues before)


ahsu1230:
lgtm, nice job!

PR #149, opened 2017-03-09
Adds Gulp Integration and Angular App Assets

qwwqwwq:


ahsu1230:
glad you're still using gulp! lolol

qwwqwwq:
Oh yeah it would be way too much work to switch that.. never had any
problem with gulp,  just with express

PR #156, opened 2017-05-18
Adds Survey Page

qwwqwwq:


ahsu1230:
Ah dang, thanks for doing this

qwwqwwq:
Turned out to be pretty easy, they support embedding

